PR #1, opened 2022-05-13
Check for the existence of `jq` before continuing

ndegruchy:
If `jq` isn't installed, the rest of the program isn't going to work. Check and bail out if it's not, telling the user what they need to do, otherwise, continue.

### DIFF
--- a/xdg-ninja.sh
+++ b/xdg-ninja.sh
@@ -7,6 +7,14 @@ then
     USE_GLOW=false
 fi
 
+# Check for the existence of jq, since it's required.
+if ! command -v jq &> /dev/null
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
Hi, thank you so much for your work, unfortunately I just fixed this myself in dac2f65809187c27832ee2c15c19d82981ebe6fe (Literally one minute before your commit).

Since it wouldn't make sense to do this check twice, I'll close this PR, but again, many thanks for your work.

ndegruchy:
Of course! I see that you posted pretty much the same fix.

Thanks!

PR #3, opened 2022-05-13
Typo fix for rbenv, add quote to solargraph

duongdominhchau:
The `$` character is missed in `rbenv` case, and the `solargraph`'s variable is not properly quoted like others suggestion.


b3nj5m1n:
Thank you so much!

PR #4, opened 2022-05-13
Improve subversion recommendation

Silejonu:


b3nj5m1n:
Thanks!

The arch wiki page is very inconsistent with this unfortunately, I tried to fix it as much as possible, but I already suspected that I missed a few.

PR #6, opened 2022-05-13
Various corrections (removed unused variables, formatted script, security and code readability changes)

Ballasi:
Thank you for your repository, it's very interesting.
I have done a few changes that I would like to push, mainly about bash code sustainability.
Most of the changes are small and insignificant (e.g., unused variables, code readability changes), but overall it'll be a good addition.
The commit messages should be self-explanatory, but if you have any question or would like to do any changes, please ask!
I tested the program on my machine with the changes, everything works fine for me.
Thanks.

P.S., I see you have a hard bash dependency, are you willing to make your program POSIX-compliant (usable for bash, zsh, dash, ...) or do you plan on keeping that hard dependency?

### DIFF
--- a/xdg-ninja.sh
+++ b/xdg-ninja.sh
@@ -80,21 +80,20 @@ log() {
     case "$MODE" in
 
     ERR)
-        printf "[\e[1;31m$NAME\e[1;0m]: \e[1;3m$FILENAME\e[1;0m\n"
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
This is awesome, thank you so much!

I personally use zsh as my shell, but use bash to run all my scripts, so I'm not really familiar with using it to run scripts. Is there any benefit if you're not using anything specific to that shell?

I'd definitely be willing to make it POSIX-compliant, though.

PR #7, opened 2022-05-13
Make the script callable in more places

traxys:
By using `/usr/bin/env bash` instead of `/bin/bash` this script can be called in distrubtions such as NixOS who don't have a /bin/bash.

By using a path relative to the script and not to the CWD we can run the script from anywhere.


b3nj5m1n:
Thanks a lot!

PR #17, opened 2022-05-14
Fixed Xresources message

Vikingat-RAGE:
Xresources message shows incorrect export variable suggestion. Fixed to correct value.


b3nj5m1n:
Thanks!

PR #19, opened 2022-05-14
README: use more generic term than bash

Ballasi:
The `README.md` files still stipulates a dependency on `bash`

I tried to use more generic terms.


b3nj5m1n:
Thank you very much!

PR #23, opened 2022-05-15
xdg-ninja: change jq json source

### LINKED ISSUE #22: Invalid string error
The script runs flawlessly after switching the shebang to bash.

The error is caused by the read on line 173.
https://github.com/b3nj5m1n/xdg-ninja/blob/91a00011d2122a3f1ebb12e45b404373ff099c27/xdg-ninja.sh#L170-L176

Terminal with "invalid string" errors

Ballasi:
There was an issue caused by inconsistencies on the implementation of echo through different shells. Some, like bash, did not interpret \n while others, like dash, did. This caused issue when sending the information to jq to parse as it as JSON specifications forbids multi-line strings.

A quick fix was applied by giving jq the path to the file to analyze rather than grabbing it from stdin.

Fixes #22

### DIFF
--- a/xdg-ninja.sh
+++ b/xdg-ninja.sh
@@ -166,14 +166,14 @@ check_file() {
 
 # Reads a file from programs/, calls check_file on each file specified for the program
 check_program() {
-    INPUT=$1
+    PROGRAM=$1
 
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
Thank you very much!

PR #24, opened 2022-05-15
Add support for bat for markdown rendering

pvonmoradi:
`bat` can render markdown to terminal. It is also more probable to be installed on user's system.
Nevertheless, this PR prioritizes `glow` for markdown rendering, if it is not found it tries to use `bat`. 

Also add a newline to log HELP to increase readability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ This will run every test in the default configuration.
 
 ### Optional
 
-- [glow](https://repology.org/project/glow/packages) for rendering markdown in the terminal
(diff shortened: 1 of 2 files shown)


b3nj5m1n:
Hi, thank you very much!

In my testing, bat's output wasn't as clear as glows, so I added a disclaimer when falling back to bat.

Merged this with 77a170c54a920a715ab92957be308814a662d8ae and 18d848aaf63e169f92602686cd59ad17c0b3c6de.

PR #27, opened 2022-05-15
Fix escape codes for POSIX `sh`

Mudskipper875:
`\e` works with shells like bash and zsh, but not with dash.
`\033` is portable and POSIX compliant.


b3nj5m1n:
Thank you so much, I wasn't aware of that!

PR #29, opened 2022-05-15
fix dirname

diogo464:
When calling xdg-ninja.sh using
`sh xdg-ninja.sh`
or
`bash xdg-ninja.sh`

the script fails with
```
jq: error: Could not open file xdg-ninja.sh/programs/*: Not a directory
jq: error: Could not open file xdg-ninja.sh/programs/*: Not a directory
Done checking your $HOME.
```

This commit just fixes that problem.


b3nj5m1n:
Thank you so much!

PR #35, opened 2022-05-16
xdg-ninja: change bat flags for retrocompatibility

### LINKED ISSUE #34: bat version compatibility
I have a small concern on bat compatibility on the program, the flag `-f` used in the bat call has been implemented in version [0.16.0](https://github.com/sharkdp/bat/releases/tag/v0.16.0) while some distributions downloads lower versions on their package manager (see https://repology.org/project/bat-cat/versions), for instance, Ubuntu 21.10 (on apt) installing version 0.12.1 of `bat`.

On my side, I don't see any difference with and without the `-f` flag, and I would believe it is the same. @pvonmoradi can you confirm?

If this is, in fact, useful, it would probably be best to check the availability of this flag.

Ballasi:
Fixes #34

### DIFF
--- a/xdg-ninja.sh
+++ b/xdg-ninja.sh
@@ -129,7 +129,7 @@ log() {
         if $USE_GLOW; then
             printf "%s\n" "$HELP" | glow -
         elif $USE_BAT; then
-            printf "%s\n" "$HELP" | bat -pp -f --language markdown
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
Thank you so much, I was just about to do this.

PR #37, opened 2022-05-16
Fix typos in .json

Willenbrink:
In some .jsons the field name was moveable instead of movable. This results in some errors as we [query for movable in xdg-ninja.sh](https://github.com/b3nj5m1n/xdg-ninja/blob/c80f3f88dc3dd768fb690d49a7ec4bfcb7cf4bc0/xdg-ninja.sh#L147).
This PR simply replaces any occurence of moveable with movable.

### DIFF
--- a/programs/aria2.json
+++ b/programs/aria2.json
@@ -3,12 +3,12 @@
     "files": [
         {
             "path": "${HOME}/.aria2/dht.dat",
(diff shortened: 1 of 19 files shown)


b3nj5m1n:
Thank you so much!

PR #40, opened 2022-05-17
sqlite: suggest XDG_CACHE_HOME instead of XDG_DATA_HOME

TalviT:
Since it is a history file, i would consider it non-essential.


b3nj5m1n:
After thinking about it a little, I think this makes sense. When I ported the arch wiki page, I set it to DATA_HOME for most history files, but the only files for which this actually makes sense are probably the shell history files.

PR #46, opened 2022-05-19
fix: update bash and zsh history file locations to use XDG_STATE_HOME

### LINKED ISSUE #45: Should shell history files be STATE, CACHE, or DATA?
For zsh/bash history files, currently:

[ArchWiki](https://wiki.archlinux.org/title/XDG_Base_Directory) suggests **STATE**.

[antidot](https://github.com/doron-cohen/antidot/blob/23cda210c9dbd18b7be3498d7a351b47f884b1aa/rules.yaml#L115) suggests **CACHE**.

xdg_ninja suggests **CACHE** for [bash](https://github.com/b3nj5m1n/xdg-ninja/blob/e05e0765bb5701d1bf772d4380c647d0ac92db1d/programs/bash.json#L7) and **DATA** for [zsh](https://github.com/b3nj5m1n/xdg-ninja/blob/e05e0765bb5701d1bf772d4380c647d0ac92db1d/programs/zsh.json#L16).

The XDG Base Directory Specification states:

>  The $XDG_STATE_HOME contains state data that should persist between (application) restarts, but that is not important or portable enough to the user that it should be stored in $XDG_DATA_HOME. It may contain:
>
>   actions history (logs, history, recently used files, …)
>
>    current state of the application that can be reused on a restart (view, layout, open files, undo history, …)

So I am leaning toward **STATE**, but figured I would ask first. :)

rocode:
Fixes #45

### DIFF
--- a/programs/bash.json
+++ b/programs/bash.json
@@ -4,7 +4,7 @@
         {
             "path": "${HOME}/.bash_history",
             "movable": true,
-            "help": "Export the following environment variables:\n\n```bash\nexport HISTFILE=${XDG_CACHE_HOME}/bash/history\n```\n"
+            "help": "Export the following environment variables:\n\n```bash\nexport HISTFILE=${XDG_STATE_HOME}/bash/history\n```\n"
         }
     ]
 }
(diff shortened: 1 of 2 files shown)


b3nj5m1n:
Thanks a lot!

PR #47, opened 2022-05-19
Add missing line ending

Kellett:
If jq is missing the script exited without a final new line

This leaves the caret at a weird place. See screenshot

Before:
image


b3nj5m1n:
Thank you, missed that when converting the echo's to printf.

PR #49, opened 2022-05-20
Double quote `$XDG...` in shell exports

Tiv0w:
This prevents globbing and word splitting.
Reference: https://github.com/koalaman/shellcheck/wiki/SC2086

I thought it was best to adjust it here, in case people copy-paste the suggestion given by `xdg-ninja` verbatim in their config.

### DIFF
--- a/programs/android-studio.json
+++ b/programs/android-studio.json
@@ -3,7 +3,7 @@
         {
             "path": "$HOME/.android",
             "movable": true,
-            "help": "Export the following environment variables:\n\n```bash\nexport ANDROID_HOME=$XDG_DATA_HOME/android\n```\n"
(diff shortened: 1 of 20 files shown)


b3nj5m1n:
Thank you so much!

We should probably standardize this at some point, since it's still a bit inconsistent.

Tiv0w:
Maybe adding it in a Contributing section of the README.
But I skimmed over the repo, and right now it seems all `$...` that need to be quoted are.
Only place where I'm not sure is the Julia config, right now it's producing a funky looking env var.

Tiv0w:
Okay, after a bit more research, it seems the current configuration of the `JULIA_DEPOT_PATH` is a bit wrong, I'll open a new PR in a minute.

PR #50, opened 2022-05-20
Remove quotes escaping for JULIA_DEPOT_PATH

Tiv0w:
Fix an incorrect quote escaping in the recommended `JULIA_DEPOT_PATH` value.

It now can be copied verbatim; following the recommendation at:
https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_DEPOT_PATH

### DIFF
--- a/programs/julia.json
+++ b/programs/julia.json
@@ -4,7 +4,7 @@
         {
             "path": "$HOME/.julia",
             "movable": true,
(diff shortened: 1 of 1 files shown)


Tiv0w:
Thanks again for the tool, it's clean and has a nice way of displaying things.

b3nj5m1n:
Thank you!

PR #51, opened 2022-05-20
Add "nvidia GLCache" and "Microsoft Visual Studio Code"

FranzXaver:
This should work as documented here, but unfortunately it does not work on my machine:
https://download.nvidia.com/XFree86/Linux-x86_64/460.67/README/openglenvvariables.html


b3nj5m1n:
Thanks!

PR #56, opened 2022-05-22
Optimize usage of jq

Layerex:
Call jq only once per file, also apply some minor refactors.
Script is about 3.5 times faster now.


b3nj5m1n:
Wow, thank you so much, this is awesome!

PR #58, opened 2022-05-23
Call jq only once and optimize string decoding

Layerex:
Script is about 14 times faster now.


b3nj5m1n:
Incredible, thank you so much!

PR #61, opened 2022-05-23
wget: correct help for hsts file

SuperSandro2000:


b3nj5m1n:
I'm not sure about this one, I think calling it history file makes it clearer what the file actually is, is there any particular reason for why you think this is better?

FranzXaver:
It is not a history file, but a HSTS (HTTP Strict Transport Security, RFC 6797) database.
From the man page:

> Wget supports HSTS (HTTP Strict Transport Security, RFC 6797) by default. ... Wget will use the supplied file as the HSTS database. Such file must conform to the correct HSTS database format used by Wget. ... The Wget's HSTS database is a plain text file. Each line contains an HSTS entry.

b3nj5m1n:
Forgive my ignorance, thank you very much!

PR #62, opened 2022-05-23
nss: explain why file may appear again

Layerex:


Layerex:
> `ln -s $XDG_DATA_HOME/pki/ ~/.pki` keeps at least the files in one place and synced.

Equally as does `rm -rf "$XDG_DATA_HOME"/pki`

b3nj5m1n:
Thank you!

Layerex:
I guess, file should be marked as immovable, since moving doesn't help

b3nj5m1n:
This is a tricky one, since it's other programs creating the file, it would be more consistent to add a duplicate entry for this file to those other programs, but that could also clutter the output.

Layerex:
> This is a tricky one, since it's other programs creating the file, it would be more consistent to add a duplicate entry for this file to those other programs, but that could also clutter the output.

Files where moving doesn't always work or is undesirable (like .zshenv) may be marked with other value of "movable" (and thus with other color)

b3nj5m1n:
I'll think about it, there's already more info available in the haskell implementation, the potential problem I see with this is that it might make things more complicated than they have to be, and I want to keep the barrier to entry as low as possible to encourage contributions. (Ik, probably shouldn't have used haskell)

Layerex:
> I'll think about it, there's already more info available in the haskell implementation, the potential problem I see with this is that it might make things more complicated than they have to be, and I want to keep the barrier to entry as low as possible to encourage contributions. (Ik, probably shouldn't have used haskell)

I don't think that adding another value of "movable" for script to interpret (for example string "problematic") in addition to true and false will introduce much hassle. Don't know about haskell implementation though

SuperSandro2000:
> > `ln -s $XDG_DATA_HOME/pki/ ~/.pki` keeps at least the files in one place and synced.
> 
> Equally as does `rm -rf "$XDG_DATA_HOME"/pki`

Well, then you are losing data.

PR #63, opened 2022-05-24
Allow setting movability of files to "problematic"

Layerex:
#62

### DIFF
--- a/programs/nss.json
+++ b/programs/nss.json
@@ -2,7 +2,7 @@
     "files": [
         {
             "path": "$HOME/.pki",
-            "movable": true,
+            "movable": "problematic",
             "help": "XDG is supported out-of-the-box, so we can simply move the file to _\"$XDG_DATA_HOME\"/pki_.\n\nBut note that some apps (chromium, for example) hardcode path to _\"$HOME\"/.pki_, so the folder may appear again, see https://bugzilla.mozilla.org/show_bug.cgi?id=818686#c11"
         }
     ],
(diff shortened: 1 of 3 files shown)


b3nj5m1n:
Sorry, but I don't think this is a good idea.

Movable is being treated as a boolean in the existing codebase, this would break a lot of it.

I think it introduces more complexity than should be necessary for this, since this is only trying to solve an edge case.

Adding an optional field of _problematic_ would be more realistic, but I still don't think these edge cases warrant that.

For now, let's set movable to false for these files, and add a disclaimer at the top or something like that. I'll edit the existing files and your other PRs.

Nonetheless, thank you very much for your work!

Layerex:
> Sorry, but I don't think this is a good idea.
> 
> Movable is being treated as a boolean in the existing codebase, this would break a lot of it.
> 
> I think it introduces more complexity than should be necessary for this, since this is only trying to solve an edge case.
> 
> Adding an optional field of _problematic_ would be more realistic, but I still don't think these edge cases warrant that.
> 
> For now, let's set movable to false for these files, and add a disclaimer at the top or something like that. I'll edit the existing files and your other PRs.
> 
> Nonetheless, thank you very much for your work!

I don't think this is an edge case, because, at least, two such problematic files are in my home directory (`.pki` and `.zshenv`), those are very common ones and there are certainly more out there. But just setting movable to false would be ok too, I guess, because the purpose of this pull request is to make files, which are actually immovable or undesirable to move, visually distinct from movable ones.

Layerex:
> Movable is being treated as a boolean in the existing codebase, this would break a lot of it.

Haskell, as a very flexible language, should be able to gracefully handle this.

b3nj5m1n:
Yes, but as I said, it would add a lot of unnecessary complexity.

PR #65, opened 2022-05-24
Add clang-format

Layerex:


b3nj5m1n:
Could you provide some additional context on this, when I run the command, the file doesn't get created, and from looking at the man page, it seems like _clang-format_ doesn't depend on this file being in _$HOME_.

b3nj5m1n:
Alright, I would suggest adding a sentence or two about how you can put the file in your project dir or any of the parent directories at the top, though. If you want to do that, go ahead, otherwise I'll do it myself in a couple of hours.

Thanks for your help!

PR #66, opened 2022-05-24
Use test for "$MOVABLE" in check_file function

Layerex:
To avoid executing it's value
Patch similar to one in #63


b3nj5m1n:
Thank you for bringing this to my attention!

PR #69, opened 2022-05-28
Use pygmentize and highlight as fallback markdown renderers

Layerex:


xdg-ninja.sh
```diff
@@ -1,16 +1,30 @@
...
+        USE_BAT=true
+        printf "Glow not found, markdown rendering will be done by bat.\n"
+    elif has_command pygmentize; then
+        printf "Nor glow, neither bat found, markdown rendering will be done by pygmentize.\n"
+        USE_PYGMENTIZE=true
+    elif has_command highlight; then
+        printf "Nor glow, nor bat, neither pygmentize found, markdown rendering will be done by highlight.\n"
+        USE_HIGHLIGHT=true
+    else
+        printf "Nor glow, nor bat, nor pygmentize, neither highlight found, markdown rendering not available.\n"
```
b3nj5m1n, line 25: I think this is getting too verbose. Also, I _think_ the order of nor and neither is incorrect, see [here](https://www.thefreedictionary.com/neither+...+nor), but I could be wrong.

I think this will suffice.
```suggestion
        printf "Markdown rendering will be done by bat. (Glow is recommended)\n"
    elif has_command pygmentize; then
        printf "Markdown rendering will be done by pygmentize. (Glow is recommended)\n"
        USE_PYGMENTIZE=true
    elif has_command highlight; then
        printf "Markdown rendering will be done by highlight. (Glow is recommended)\n"
        USE_HIGHLIGHT=true
    else
        printf "Markdown rendering not available. (Glow is recommended)\n"
```

PR #70, opened 2022-05-28
Normalize number of trailing newlines in markdown to 2

Layerex:
For output to look uniform with formatters like bat


b3nj5m1n:
Nice, thank you!

PR #71, opened 2022-05-30
Update cabal and opam description

Willenbrink:
The cabal help advises to move `CABAL_DIR` to XDG_CACHE_HOME. I think it should point to XDG_DATA_HOME as that dir includes the installed binaries and they represent data, not a cache in my mind.

This PR also expands the opam help as I had some problems with getting it to work. Although I haven't tested that these instructions alone are enough, they should help people diagnose their own problems and lookup help online.

### DIFF
--- a/programs/cabal.json
+++ b/programs/cabal.json
@@ -3,7 +3,7 @@
         {
             "path": "$HOME/.cabal",
             "movable": true,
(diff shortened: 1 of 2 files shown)


b3nj5m1n:
Thanks!

Issue #80, opened 2022-06-04
List of missing programs


a2n-s:
I really love that application :star_struck: 
I was able to quickly clean a lot of my `$HOME` directory and my dotfiles repo, really cool stuff we have here :+1: 

I wanted to further clean my `$HOME` and noticed, and that's normal considering the number of existing applications, a lot of hidden files and directories still hanging around without explicit permissions :thinking: 

Below are lists of all the directories and files I still have in my `$HOME`, I only removed those already supported by `xdg-ninja` but without definitive solution and my personal files. There might be items not relevant to `xdg-ninja`, please tell me and i'll remove them :wink:

If you prefer to split this issue into more little ones, this one can be closed and the job can be done.

I just thought a somewhat ""complete"" list of missing programs would be usefull :yum: 
And I would be more than happy to help if you accept external contributions :yum: 

**Missing directories:**
```
.arduino-create/
.arduino15/
.dwarffortress/
.eendroroy-coloschemes/
.emacs.d/
.gsutil/
.ipython/
.keras/
.kube/
.links/
.lyrics/
.moc/
.mujoco/
.oh-my-bash/
.oh-my-xonsh/
.pytest_cache/
.quickemu/
.screenlayout/
.slime/
.SpaceVim/
.SpaceVim.d/
.surf/
.telegram-cli/
.themes/
.thumbnails/
.trash/
.vim/ -> ~/.SpaceVim/
.vim_back/
.virtualbox/
.when/
.xmonad/
.zotero/
```

**Missing files:**
```
.bash_aliases
.bash_logout
.bash_profile
.bashrc
.bashrc.backup
.boto
.ctags
.dmrc
.gcalcli_cache
.gcalcli_oauth
.gnuplot_history
.grip-oggenc
.luahistory
.matho_history
.net.log
.notion-enhancer
.nwinkler_random_colors
.osh-update
.pdbhistory
.profile
.pulse-cookie
.python_history-83827.tmp
.rtorrent.rc
.sc_history
.sdirs
.viminfo
.vimrc_back
.xonshrc
.xscreensaver
.yarnrc
.zcompdump*
.zsh_history
```
**Special cases**:
- `gnupg`: this program is supported but the `~/.gnupg` file gets regenerated each time on my machine. This would be great to have a mention to that problem in the output of `xdg-ninja`

a2n-s:
I removed `~/.steampid` as it is mentionned in #78.

wael444:
[...]

PR #81, opened 2022-06-05
Address program list from issue #80

a2n-s:
Are you ok with that start on the work we discussed in #80, @b3nj5m1n?

I thought that, if there are things to change, the earlier the better!


b3nj5m1n:
Yes, this is awesome, thank you so much!

PR #95, opened 2022-06-09
Modifications that improve scriptability of code

### LINKED ISSUE #73: Feature request: dump results to a text file
I'd love to see a command-line option to show the results in plain text, i.e., no color control codes or markdown, so I could redirect it to a text file.

### LINKED ISSUE #93: Add option to hide unsupported files and directories
As the title says. This would be useful for a couple of different uses cases, such as:
- Users who use the command in `watch` to get live updates.
- Users who want a de-cluttered output.

### LINKED ISSUE #94: Exit status dependent upon amount of fixable files in homedir
This would allow for slightly easier scripting using the program.

Example of a cronjob that could be created using this feature:
```sh
@weekly xdg-ninja.sh || notify-send "Unnecessary dot files in homedir" "You have $@ unnecessary files in your home directory."
```

Currently a user would have to create a rather-long wrapper program that finds the amount of red ansi codes which is not preferable.

aflyingpumpkin:
I apologize for the rather-large scope of this merge, if there are any particular features that are not suitable for this project let me know and I can remove them from the merge request.

Features added:
- Changed help parsing to use case instead of if statements.
- Intelligent color detection, and ``--color=` flag
  - Do not apply color when being piped into a TTY.
  - Do not apply color if user has the "$NO_COLOR" environment variable set.
  - ``--color=` flag which acts in the same way it does for programs such as `ls` with `auto` (default), `always`, and `never` options.
- Added ``--decoder=` flag which allows users to define their own. (These two features should close #73)
- Flag for json output to allow for the user to parse the output in a shell script more easily.
- Flag to skip warnings to clean up output, which closes #93
- exit code dependent upon amount of ERRs logged, which closes #94


b3nj5m1n:
I'm quite happy with most of the changes you've made here, but there are a few issues.

I won't have time to properly look into all of this now, I'll probably get to it tomorrow.

Here's what I've noticed so far:
- The checks for missing XDG vars don't output colours (I get a literal `{ANSI_BEGIN}${ANSI_BOLD}${ANSI_CYAN}${ANSI_END}The $XDG_STATE_HOME environment variable is not set, make sure to add it to your shell's configuration before setting any of the other environment variables!${ANSI_BEGIN}${ANSI_BOLD}${ANSI_CLEAR}${ANSI_END}` in the output)
- The json output option outputs multiple json objects, instead of one large json array. This is not something you necessarily have to take care of, having it like this is better than not having it at all, but it should make it easier to process this output.
- (There probably shouldn't be any other output when passing the json flag. Again, better like this than not at all, but would make it easier to process the output.) _Since this would require even more fundamental changes to the script, let's not address it in this PR._
- I'm not sure about how you're handling ansi codes, it makes it easier to read for those of us who haven't memorized the codes, but I think there might be better ways to do this. This is also something you don't have to handle directly for this to be merged, but the shellcheck warnings do worry me, even though it should be fine.

Overall, this is really cool, thank you! I'll get back to this tomorrow.

xdg-ninja.sh
```diff
@@ -5,83 +5,126 @@ has_command() {
...
+
+help() {
+    set_colors
+    HELPSTRING="""\
+
+
+        ${ANSI_BEGIN}${ANSI_BOLD}${ANSI_WHITE}${ANSI_BACKGROUND_PURPLE}${ANSI_END}xdg-ninja${ANSI_BEGIN}${ANSI_CLEAR}${ANSI_END}
 
-    \033[1;3mCheck your \$HOME for unwanted files.\033[1;0m
+        ${ANSI_BEGIN}${ANSI_BOLD}${ANSI_ITALLIC}${ANSI_END}Check your \$HOME for unwanted files.${ANSI_BEGIN}${ANSI_BOLD}${ANSI_CLEAR}${ANSI_END}
```
Layerex, line 68: Word italic is written with one l

xdg-ninja.sh
```diff
@@ -5,22 +5,20 @@ has_command() {
     return $?
 }
 
-USE_GLOW=false
-USE_BAT=false
-USE_PYGMENTIZE=false
-USE_HIGHLIGHT=false
+DECODER="cat"
```
Layerex, line 8: Why use `cat`? Maybe, just don't pipe, if there is no decoder

xdg-ninja.sh
```diff
@@ -30,58 +28,115 @@ fi
 
 unalias -a
 
-HELPSTRING="""\
+#Set ansi color code variables
+set_colors() {
+    case $COLOR in
```
Layerex, line 33: I think, `always` option should be added for completeness

xdg-ninja.sh
```diff
@@ -30,58 +28,115 @@ fi
...
-        printf "%b" "$HELPSTRING"
-        exit
-    elif [ "$i" = "--skip-ok" ]; then
-        SKIP_OK=true
-    elif [ "$i" = "--no-skip-ok" ]; then
-        SKIP_OK=false
-    elif [ "$i" = "-v" ]; then
-        SKIP_OK=false
-    fi
+    case $i in
```
Layerex, line 94: Add an option to provide arguments without `=` (`--decoder x`, not only `--decoder=x`)

xdg-ninja.sh
```diff
@@ -196,21 +245,21 @@ check_file() {
 # Reads files from programs/, calls check_file on each file specified for each program
 do_check_programs() {
     while IFS="
-" read -r name; read -r filename; read -r movable; read -r help; do
-        check_file "$name" "$filename" "$movable" "$help"
+" read -r name; read -r filename; read -r movable; read -r help; read -r json_file;  do
+        check_file "$name" "$filename" "$movable" "$help" "$json_file"
     done <<EOF
-$(jq 'inputs as $input | $input.files[] as $file | $input.name, $file.path, $file.movable, $file.help' "$(dirname "$0")"/programs/* | sed -e 's/^"//' -e 's/"$//')
+$(jq 'inputs as $input | $input.files[] as $file | $input.name, $file.path, $file.movable, $file.help, input_filename' "$(dirname "$0")"/programs/* | sed -e 's/^"//' -e 's/"$//')
```
Layerex, line 251: Why parse json files with slow jq, if all you are going to do is print those?

Layerex, line 251: Also, what about joining json files into a single json for more convenient parsing (without jq or with it)?

xdg-ninja.sh
```diff
@@ -30,58 +28,115 @@ fi
...
+}
 
+help() {
+    set_colors
+    HELPSTRING="""\
 
-    \033[37;45;1mxdg-ninja\033[0m
 
-    \033[1;3mCheck your \$HOME for unwanted files.\033[1;0m
+        ${ANSI_BEGIN}${ANSI_BOLD}${ANSI_WHITE}${ANSI_BACKGROUND_PURPLE}${ANSI_END}xdg-ninja${ANSI_BEGIN}${ANSI_CLEAR}${ANSI_END}
```
Layerex, line 66: That escape code composition looks really ugly and litters the strings. Is it really needed there? What about just defining all escape codes used (like `BOLD_ITALIC`, `CLEAR_BOLD`, `CLEAR`, `CYAN_BOLD`, etc.) however (with composition or without) first and using those instead

xdg-ninja.sh
```diff
@@ -30,58 +28,115 @@ fi
...
+        --color=*)
+            COLOR="${i#*=}"
+            ;;
+        --help|-h)
+            help
+            ;;
+        --skip-ok)
+            SKIP_OK=true
+            ;;
+        --skip-warn)
```
Layerex, line 104: Also, `--no-skip-warn` option should be added

xdg-ninja.sh
```diff
@@ -30,58 +28,115 @@ fi
...
+            SKIP_WARN=true
+            ;;
+        --no-skip-ok|-v)
+            SKIP_OK=false
+            ;;
+        --decoder=*)
+            DECODER="${i#*=}"
+            ;;
+        --json)
```
Layerex, line 113: And an option for disabling json output, maybe (`--render`, `--human-readable`?)

xdg-ninja.sh
```diff
@@ -5,22 +5,20 @@ has_command() {
...
+    DECODER="glow -"
 else
     if has_command bat; then
-        USE_BAT=true
+        DECODER="bat -pp --decorations=always --color=always --language markdown"
         printf "Markdown rendering will be done by bat. (Glow is recommended)\n"
     elif has_command pygmentize; then
+        DECODER="pygmentize -l markdown"
         printf "Markdown rendering will be done by pygmentize. (Glow is recommended)\n"
         USE_PYGMENTIZE=true
```
Layerex, line 18: The variable is now unused and should be removed

Layerex (changes requested):

xdg-ninja.sh
aflyingpumpkin: I considered that, though to not pipe would require adding in an if statement which makes this much less elegant.

xdg-ninja.sh
aflyingpumpkin: If `always` was added it would do nothing, or I would have to move all of the variable setting inside of `auto` and `always`, and add a catchall that outputs an error which will increase line-count and make stuff less readable.

xdg-ninja.sh
aflyingpumpkin: The files need to be parsed by `jq` for `xdg-ninja.sh` to check if the file exists or not.

aflyingpumpkin: A `-q|--quite` option is also needed for more convenient parsing.

xdg-ninja.sh
Layerex: Writing `DECODER="cat"` instead of `DECODER=` is much less elegant. Also, spawning external processes for no reason is slow

xdg-ninja.sh
Layerex: No. Globs don't output non-existent files, you know

Layerex: `--quiet`? I don't get it

aflyingpumpkin: All files are existent in `./programs/*` that is just a database of the possible files that could exist in ones homedir.

aflyingpumpkin: Currently if you use the `--json` flag the program first shills glow (if uninstalled), declares it is checking the homedir, outputs the json, says it is done checking the homedir, then asks for you to contribute to the github. All of this is also very inconvenient for people who wish to parse the output.

xdg-ninja.sh
Layerex: Do nothing then, and warn user if `COLOR` is not in {`never`, `auto`, `always`}. Also add `always` option to help

Layerex: I also suggest to prefix environment variable names with `XDG_NINJA_`, since variable names like `COLOR`, `DECODER` are pretty common and may mess things up if set globally

aflyingpumpkin: `NO_COLOR` is actually a standard:
https://no-color.org/

xdg-ninja.sh
Layerex: > All files are existent in `./programs/*` that is just a database of the possible files that could exist in ones homedir.

Sorry, misunderstood you

xdg-ninja.sh
aflyingpumpkin: The current way arg checking is setup does not support this, without looping through every single argument twice if `--decoder` is present.

xdg-ninja.sh
Layerex: Do it. But don't forget about joining json files. With current file format it is easy: you output `,` after each file and surround output with `[` and `]`

aflyingpumpkin: Oh ok, I thought I would have to store all the jsons into a file, then run that file through `jq` at the end of the program.

xdg-ninja.sh
```diff
@@ -168,19 +212,24 @@ check_file() {
     FILENAME="$2"
     MOVABLE="$3"
     HELP="$4"
+    JSON_FILE="$5"
```
Layerex, line 215: I think, there should be separate `check_file` function for json files, since output format is totally different

Layerex (changes requested):

xdg-ninja.sh
Layerex: > `NO_COLOR` is actually a standard: https://no-color.org/

Ok, don't change it then, but deal with other environment variable names

xdg-ninja.sh
Layerex: > The current way arg checking is setup does not support this, without looping through every single argument twice if `--decoder` is present.

Then change it. You don't have to loop more than one time for that, though.

xdg-ninja.sh
Layerex: > Oh ok, I thought I would have to store all the jsons into a file, then run that file through `jq` at the end of the program.

Nevermind, my solution won't work, if you need to output files, not programs (take a look at json files). Asking `jq` to provide jsons for files and then reprinting those as I suggested would work

xdg-ninja.sh
Layerex: `--text`?

aflyingpumpkin: I'll just go with `--no-json` for simplicity sake

xdg-ninja.sh
aflyingpumpkin: That's the only environment variable the program reads (other than the XDG ones)

Layerex: Isn't `COLOR` read? I also would let user to provide one's own decoder.

xdg-ninja.sh
Layerex: > I'll just go with `--no-json` for simplicity sake

But if there will be other output modes?

xdg-ninja.sh
aflyingpumpkin: Nope, none of that is read from env variables, just use flags. I could pretty easily implement a config file in this PR ~~as well ||and make it default to `~/.xdg-ninjarc`.~~

xdg-ninja.sh
Layerex: > But if there will be other output modes?

Those probably won't, but I nevertheless think `--text` fits better here

xdg-ninja.sh
Layerex: > Nope, none of that is read from env variables, just use flags. I could pretty easily implement a config file in this PR ~as well ||and make it default to `~/.xdg-ninjarc`.~

:)

So this is resolved

aflyingpumpkin:
Another current issue with this MR is that most of the options I added are dependent on order.
```
xdg-ninja --help --color=never
```
Will not work as expected.

xdg-ninja.sh
aflyingpumpkin: While the output format is totally different most of the logic is still the exact same.

aflyingpumpkin:
Sorry about that, I'll undo that commit.

aflyingpumpkin:
All issues mentioned should be resolved now

xdg-ninja.sh
```diff
@@ -5,91 +5,172 @@ has_command() {
...
+        ${ANSI_ITALIC}-h${ANSI_CLEAR}
 
-    ────────────────────────────────────
+        ${ANSI_ITALIC}--skip-ok${ANSI_CLEAR}            ${ANSI_BOLD}Don't display anything for files that do not exist (default)${ANSI_CLEAR}
+        ${ANSI_ITALIC}--skip-warn${ANSI_CLEAR}          ${ANSI_BOLD}Don't display anything for files that cannot be fixed.${ANSI_CLEAR}
+        ${ANSI_ITALIC}--no-skip-ok${ANSI_CLEAR}         ${ANSI_BOLD}Display messages for all files checked${ANSI_CLEAR}
+        ${ANSI_ITALIC}--no-skip-warn${ANSI_CLEAR}       ${ANSI_BOLD}Don't display anything for files that cannot be fixed.${ANSI_CLEAR}
+        ${ANSI_ITALIC}--color=WHEN${ANSI_CLEAR}         ${ANSI_BOLD}Color the output always, never, or auto (default)${ANSI_CLEAR}
+        ${ANSI_ITALIC}--decoder=DECODER${ANSI_CLEAR}    ${ANSI_BOLD}Manually set the decoder used for markdown.${ANSI_CLEAR}
+        ${ANSI_ITALIC}--output-style=STYLE${ANSI_CLEAR} ${ANSI_BOLD}Style of ouptut, either normal (defualt) or json${ANSI_CLEAR}
```
Layerex, line 59: Typos

xdg-ninja.sh
```diff
@@ -5,91 +5,172 @@ has_command() {
...
 
-    \033[1;3mCheck your \$HOME for unwanted files.\033[1;0m
+        ${ANSI_ITALIC}--help${ANSI_CLEAR}               ${ANSI_BOLD}This help menu${ANSI_CLEAR}
+        ${ANSI_ITALIC}-h${ANSI_CLEAR}
 
-    ────────────────────────────────────
+        ${ANSI_ITALIC}--skip-ok${ANSI_CLEAR}            ${ANSI_BOLD}Don't display anything for files that do not exist (default)${ANSI_CLEAR}
+        ${ANSI_ITALIC}--skip-warn${ANSI_CLEAR}          ${ANSI_BOLD}Don't display anything for files that cannot be fixed.${ANSI_CLEAR}
+        ${ANSI_ITALIC}--no-skip-ok${ANSI_CLEAR}         ${ANSI_BOLD}Display messages for all files checked${ANSI_CLEAR}
+        ${ANSI_ITALIC}--no-skip-warn${ANSI_CLEAR}       ${ANSI_BOLD}Don't display anything for files that cannot be fixed.${ANSI_CLEAR}
```
Layerex, line 56: Duplicates `--skip-warn`

xdg-ninja.sh
```diff
@@ -5,91 +5,172 @@ has_command() {
...
-        printf "%b" "$HELPSTRING"
-        exit
-    elif [ "$i" = "--skip-ok" ]; then
-        SKIP_OK=true
-    elif [ "$i" = "--no-skip-ok" ]; then
-        SKIP_OK=false
-    elif [ "$i" = "-v" ]; then
-        SKIP_OK=false
-    fi
+    case $i in
```
Layerex, line 82: You still haven't implemented proper argument parsing

xdg-ninja.sh
```diff
@@ -195,23 +271,26 @@ check_file() {
 
 # Reads files from programs/, calls check_file on each file specified for each program
 do_check_programs() {
+    [ "$OUTPUT_STYLE" = "json" ] && mkdir "$XDG_RUNTIME_DIR"/xdg-ninja/
```
Layerex, line 274: Logic of outputting json should be decoupled

Layerex: > While the output format is totally different most of the logic is still the exact same.

Then why are `[ "$OUTPUT_STYLE" = "json" ]` tests are everywhere?

xdg-ninja.sh
```diff
@@ -195,23 +271,26 @@ check_file() {
...
     while IFS="
-" read -r name; read -r filename; read -r movable; read -r help; do
-        check_file "$name" "$filename" "$movable" "$help"
+" read -r name; read -r filename; read -r movable; read -r help; read -r json_file;  do
+        check_file "$name" "$filename" "$movable" "$help" "$json_file"
     done <<EOF
-$(jq 'inputs as $input | $input.files[] as $file | $input.name, $file.path, $file.movable, $file.help' "$(dirname "$0")"/programs/* | sed -e 's/^"//' -e 's/"$//')
+$(jq 'inputs as $input | $input.files[] as $file | $input.name, $file.path, $file.movable, $file.help, input_filename' "$(dirname "$0")"/programs/* | sed -e 's/^"//' -e 's/"$//')
 EOF
+    [ "$OUTPUT_STYLE" = "json" ] && jq $JQ_COLOR_VAR -s . "$XDG_RUNTIME_DIR"/xdg-ninja/* && rm -rf "$XDG_RUNTIME_DIR/xdg-ninja"
```
Layerex, line 281: Quote `JQ_COLOR_VAR` maybe?

xdg-ninja.sh
```diff
@@ -168,19 +238,25 @@ check_file() {
...
 
     0)
+        [ "$SKIP_OK" = false ] && [ "$OUTPUT_STYLE" = "json" ] && cat "$JSON_FILE" && return
         log SUCS "$NAME" "$FILENAME" "$HELP"
         ;;
 
     1)
         if [ "$MOVABLE" = true ]; then
+            EXIT_STATUS=$(expr $EXIT_STATUS + 1)
+            [ "$OUTPUT_STYLE" = "json" ] && cat "$JSON_FILE" > "$XDG_RUNTIME_DIR"/xdg-ninja"$NAME".json && return
```
Layerex, line 255: You are doing it wrong. Many files of the same program may be addressed is one JSON file. JSON files are for *programs* not for *files*

Layerex (changes requested):

Layerex:
> MR now closes #94, (I will fix all of the minor bugs in a single commit once I have all the major stuff out of the way,)

In theory, there may be more than 255 files

b3nj5m1n:
> > MR now closes #94, (I will fix all of the minor bugs in a single commit once I have all the major stuff out of the way,)
> 
> In theory, there may be more than 255 files

I think we can just ignore that and return 255 for 255 or more files.

xdg-ninja.sh
aflyingpumpkin: I did... It was rejected.

xdg-ninja.sh
aflyingpumpkin: It's always just -M so it does not require a quote.

xdg-ninja.sh
aflyingpumpkin: There are 5, this is FAR less then the amount of code that would be added by splitting each into their own functions.

Layerex: It is a lot. Those make code messy and unreadable

Layerex:
> > In theory, there may be more than 255 files
> 
> I think we can just ignore that and return 255 for 255 or more files.

Yes

xdg-ninja.sh
Layerex: Doing like that is somewhat of a bad practice

xdg-ninja.sh
Layerex: Oh, ok

xdg-ninja.sh
b3nj5m1n: You can keep it as is using cat.

b3nj5m1n:
I think the scope of this PR is far too large to be properly reviewable, so I switched to using variables for ANSI codes in 79e71eaef3ab8cb2e8ec479a54f735c5b19056e2.

I'd also propose that we stick to the current version of the argument parsing using if statements for now, I'm open to addressing this later.

That should make this much more manageable, and hopefully we'll get it merged soon.

I'll see if I can resolve the merge conflicts later myself. Hope you're on board with this, if you have any thoughts let me know.

b3nj5m1n:
I fixed the merge conflicts, found a few issues along the way.

- The decoder option doesn't seem to do anything
- The json output option outputs the entire json file, even though some files aren't actually present, I believe @Layerex had already pointed this out too
- The json output option emits errors at the beginning (second one is only present if it's been run before):
```
mkdir: cannot create directory ‘/run/user/1000/xdg-ninja/’: File exists
Input error: Is a directory
```

---

I'm beginning to think it might be better to split this up into multiple PRs, since I think this introduces more complexity than is necessary.

Let me know what you think.

Layerex:
> * The json output option emits errors at the beginning (second one is only present if it's been run before):

`mkdir -p` should be used.

JSON output should be done completely differently. Writing to file is redundant.

xdg-ninja.sh
```diff
@@ -278,20 +299,25 @@ do_check_programs() {
...
+        printf "\n"
+    fi
 }
 
 
+auto_set_decoder
+set_colors
 check_programs
-exit $EXIT_STATUS
+exit "$EXIT_STATUS"
```
Layerex, line 323:
```bash
if [ "$EXIT_STATUS" -gt 255 ]; then
    EXIT_STATUS=255
fi
```
should be added

xdg-ninja.sh
```diff
@@ -278,20 +299,25 @@ do_check_programs() {
...
-# sed is to trim quotes
+    [ "$OUTPUT_STYLE" = "json" ] && jq "$JQ_COLOR_VAR" -s . "$XDG_RUNTIME_DIR"/xdg-ninja/* && rm -rf "$XDG_RUNTIME_DIR/xdg-ninja"
 }
 
 check_programs() {
-    loudprintf "${ANSI_BOLD_ITALIC}Starting to check your ${ANSI_BOLD_CYAN}\$HOME.${ANSI_BOLD_CLEAR}\n"
-    loudprintf "\n"
+    if [ "$QUIET" != false ]; then
+        printf "%bStarting to check your %b\$HOME%b.\n" "${FX_BOLD}${FX_ITALIC}" "${FG_CYAN}" "${FX_RESET}"
+        printf "\n"
```
Layerex, line 308: Why use multiple `printf`s there?

Layerex, line 308: Alias solution was more elegant, in my opinion

xdg-ninja.sh
```diff
@@ -124,53 +109,89 @@ auto_set_decoder() {
...
+    printf '%b%s%b\n' "${FX_BOLD}${FG_CYAN}" "The \$XDG_CACHE_HOME environment variable is not set, make sure to add it to your shell's configuration before setting any of the other environment variables!" "${FX_RESET}"
+    printf "%b    ⤷ The recommended value is: %b\$HOME/.cache%b\n" "${FX_BOLD}${FG_CYAN}" "${FX_BOLD}${FX_ITALIC}" "${FX_RESET}"
 fi
-if [ -z "${XDG_RUNTIME_DIR}" ]; then
-    XDG_RUNTIME_DIR=/tmp/
-    loudprintf "${ANSI_BOLD_CYAN}%s${ANSI_BOLD_CLEAR}\n" "The \$XDG_RUNTIME_DIR environment variable is not set, make sure to add it to your shell's configuration before setting any of the other environment variables!"
-    loudprintf "${ANSI_BOLD_CYAN}    ⤷ ${ANSI_BOLD}The recommended value is: ${ANSI_BOLD_ITALIC}/run/user/\$UID${ANSI_BOLD_CLEAR}\n"
+if [ -z "${XDG_RUNTIME_DIR}" ] && [ "$QUIET" != false ]; then
+    printf '%b%s%b\n' "${FX_BOLD}${FG_CYAN}" "The \$XDG_RUNTIME_DIR environment variable is not set, make sure to add it to your shell's configuration before setting any of the other environment variables!" "${FX_RESET}"
+    printf "%b    ⤷ The recommended value is: %b/run/user/\$UID%b\n" "${FX_BOLD}${FG_CYAN}" "${FX_BOLD}${FX_ITALIC}" "${FX_RESET}"
```
Layerex, line 184: Maybe use XDG variable name there?

Layerex: And in other `printf`s

xdg-ninja.sh
```diff
@@ -124,53 +109,89 @@ auto_set_decoder() {
...
+        "always")
+            return
+            ;;
+        "auto")
+            if [ -t 1 ] && [ "$NO_COLOR" != false ]; then # Check if used in a pipe or if the NO_COLOR env variable is set.
+                return
+            fi
+            ;;
+    esac
+    DECODER="cat"
```
Layerex, line 148: As I said, cat shouldn't be used as default decoder, since external processes shouldn't be spawned without a need

Layerex (changes requested):

PR #97, opened 2022-06-10
Set movability of .zshenv to false, since fix is undesirable

Layerex:


b3nj5m1n:
Why do you think this is undesirable? I personally use this without issue.

Layerex:
> Why do you think this is undesirable? I personally use this without issue.

- Editing .zshenv becomes problematic
- It sets environment variables for all users, which may be undesirable

b3nj5m1n:
I don't think this warrants setting the movability to false, but we could add a disclaimer.

Layerex:
> I don't think this warrants setting the movability to false, but we could add a disclaimer.

I think, since there is no value like `problematic` for movability, all files, moving which may be unwanted, should be marked as immovable

Layerex:
> * Editing .zshenv becomes problematic
> * It sets environment variables for all users, which may be undesirable

Specifically for .zshenv there is a better fix, which doesn't have those drawbacks. I will contribute it later

b3nj5m1n:
> I think, since there is no value like problematic for movability, all files, moving which may be unwanted, should be marked as immovable

I disagree, especially in this case, it's moveable, and it should be up to the user to decide whether they want to configure it system-wide, and I don't think configuring it system-wide is an issue, since I can't think of a reason my you would want the zsh files in the home directory for another user.

> Specifically for .zshenv there is a better fix, which doesn't have those drawbacks. I will contribute it later

Curious to see this, when I originally moved my zsh files, this was the only solution that worked.

Layerex:
> I think, since there is no value like problematic for movability, all files, moving which may be unwanted, should be marked as immovable
This is to user can visually distinguish files one decided not to move

PR #102, opened 2022-06-17
Added name to zoom.json

LukasPietzschmann:
`zoom.json` was missing the `name` attribute, which caused "[null]" to show up as a name in the terminal. This PR fixes that.


b3nj5m1n:
Thank you!

PR #103, opened 2022-06-21
Added GDB and GEF

LukasPietzschmann:
The env. var. for GDB is documented here: https://sourceware.org/gdb/current/onlinedocs/gdb/Command-History.html
The env. var. for GEF is not officially documented, but I found it in the source code: https://github.com/hugsy/gef/blob/f8905794baa2ac6cd1bcaee8c7b0e164f89f18dd/gef.py#L134
The fact that GEF overwrites GDBHISTFILE is also not documented. I found that in the source code, too: https://github.com/hugsy/gef/blob/f8905794baa2ac6cd1bcaee8c7b0e164f89f18dd/gef.py#L10761

Note: There already exists a gdb.json file, but there is a cleaner
method that the one proposed there, so I added the cleaner one, too.


b3nj5m1n:
Actually, I just checked the existing config for gdb, and it seems that it targets a different file?

If there are two different files created by gdb, obviously don't overwrite the config for the existing one. But since I don't use gdb myself yet I think you'll probably know better what to do here.

LukasPietzschmann:
> Actually, I just checked the existing config for gdb, and it seems that it targets a different file?

Absolutely! I don't know why I didn't noticed that ':)


I moved the gdb part of my PR to the original gdb.json file and renamed the other one to jef.json, as you suggested.

b3nj5m1n:
Thanks a lot!

PR #109, opened 2022-06-28
Add FreeCAD config

adrianinsaval:
In the .FreeCAD directory a lot of cached stuff is also present, shown selected here, do these need to be mentioned too?
image

### DIFF
--- a/programs/FreeCAD.json
+++ b/programs/FreeCAD.json
@@ -1,0 +1,30 @@
+{
+    "files": [
+        {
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
First of all, thanks for you contribution.

---

> In the .FreeCAD directory a lot of cached stuff is also present, shown selected here, do these need to be mentioned too?

Yes, everything that is present in the home directory should be mentioned. Feel free to open another PR adding these files.

PR #112, opened 2022-06-28
Added config for mcfly

HimDek:
Reference: https://github.com/cantino/mcfly#database-location

### DIFF
--- a/programs/mcfly.json
+++ b/programs/mcfly.json
@@ -1,0 +1,11 @@
+
+{
+    "files": [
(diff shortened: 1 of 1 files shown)


programs/mcfly.json
```diff
@@ -0,0 +1,11 @@
+
+{
```
b3nj5m1n, line 2:
```suggestion
{
```

b3nj5m1n:
Thanks!

PR #113, opened 2022-06-28
Added config for oh-my-zsh

HimDek:
Reference: https://github.com/ohmyzsh/ohmyzsh#custom-directory
[Relevant issue](https://github.com/ohmyzsh/ohmyzsh/issues/9543)

### DIFF
--- a/programs/ohmyzsh.json
+++ b/programs/ohmyzsh.json
@@ -4,7 +4,7 @@
         {
             "path": "$HOME/.oh-my-zsh",
             "movable": true,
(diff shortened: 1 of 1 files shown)


programs/ohmyzsh.json
```diff
@@ -0,0 +1,11 @@
+
+{
```
b3nj5m1n, line 2:
```suggestion
{
```

b3nj5m1n:
Thank you!

PR #117, opened 2022-06-28
XCURSOR_PATH should be set for custom cursors

polirritmico:
Added an export instruction for cursor themes installed in ${XDG_DATA_HOME}/icons

If the user has a custom cursor theme installed there may be problems moving $HOME/.icons to $XDG_DATA_HOME/icons if the path is not in XCURSOR_PATH. More info in the [ArchWiki](https://wiki.archlinux.org/title/Cursor_themes#Environment_variable).

### DIFF
--- a/programs/icons.json
+++ b/programs/icons.json
@@ -4,7 +4,7 @@
         {
             "path": "${HOME}/.icons",
             "movable": true,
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
Thank you!

PR #118, opened 2022-06-29
Added config for fltk

jopejoe1:
Reference: https://www.fltk.org/str.php?L3370+P0+S0+C0+I0+E0+V%25+Qxdg and https://github.com/fltk/fltk/issues/303

### DIFF
--- a/programs/fltk.json
+++ b/programs/fltk.json
@@ -1,0 +1,10 @@
+{
+    "files": [
+        {
+            "path": "$HOME/.fltk",
+            "movable": false,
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
Thanks!

Darukutsu:
I maybe don't understand but from [this merge]( https://github.com/fltk/fltk/pull/377) I think it's movable. Also see that [original issue ](https://github.com/fltk/fltk/issues/79) is solved.

jopejoe1:
The latest realese was before that pr was merged, so it not in the current realese, but it will most likely be in the next realese.

PR #123, opened 2022-06-29
Added config for vscode and code-oss

HimDek:
Reference: https://github.com/microsoft/vscode/issues/3884


b3nj5m1n:
Thanks!

HimDek:
The name of the open sorce variant is code or in some cases code-oss and not vscode.

PR #124, opened 2022-06-30
Fix `.android` info

HanabishiRecca:
Seems like existing info is wrong.


b3nj5m1n:
Hi, sorry for taking so long to get back to you, I don't think it's a good idea to just remove existing info, even if it only partially works, so I added it as a disclaimer at the end.

Thanks for your contribution!

PR #128, opened 2022-07-05
Some minor changes to README

0WN463:
* Fix the formatting of $HOME
* Use "Markdown" instead of "markdown", as that is the proper casing for the term


b3nj5m1n:
Thanks!

PR #134, opened 2022-07-10
Add `Flatpak` (`.var`) info

HanabishiRecca:


b3nj5m1n:
Hi, is _.var_ actually only used by flatpak? I.e. is the directory structure _~/.var/files_ or more like _~/.var/flatpak/files_?

I haven't been able to find much about this online, but [this](https://askubuntu.com/a/1353030) answer concerns me a little.

b3nj5m1n:
Thanks for the link, should be safe to merge.

PR #136, opened 2022-07-11
Add visidata config

fbearoff:
Add a config for visidata which supports XDG since https://github.com/saulpw/visidata/pull/1420


b3nj5m1n:
Thanks very much!

PR #145, opened 2022-07-23
Added mutt-wizard isync environment variable

Riccardiunix:
Added environment variable for isync in the case it is used by [mutt-wizard](https://github.com/LukeSmithxyz/mutt-wizard)
The name of the envirorment varible is hardcoded in the [mw](https://github.com/LukeSmithxyz/mutt-wizard/blob/master/bin/mw) file


b3nj5m1n:
Thanks!

PR #146, opened 2022-07-24
Fix .cargo file to be movable

hizani:
Idk why cargo's `movable` is set to false [here](https://github.com/b3nj5m1n/xdg-ninja/commit/0ed6693f66fb949929b6d59b1b79dc87be538d61) with `help` saying it's movable. 
Works for me.


b3nj5m1n:
Jupp, that's a mistake, thanks for fixing!

PR #150, opened 2022-08-06
Add support for rofi's .rofi.pid

surrealoatmeal:
[rofi](https://github.com/davatorium/rofi) creates a pid file in `$HOME` to keep track of instances. Moving it is possible via the `-pid` argument. This the implementation thereof.

### DIFF
--- a/programs/renpy.json
+++ b/programs/renpy.json
@@ -1,0 +1,10 @@
+{
+    "name": "Ren'Py",
+    "files": [
+        {
+            "path": "$HOME/.renpy",
+            "movable": false,
+            "help": "Not going to be supported. See the relevant [issue](https://github.com/renpy/renpy/issues/1377).\n\nThere's a workaround in the comments that needs to be applied to every game."
+        }
+    ]
+}


b3nj5m1n:
Hi, first of all, thank you for your contribution!

I use rofi, and I've never seen this file, do you happen to know if is it only created under specific circumstances, or if it's something from an older version?

surrealoatmeal:
Looking at rofi's source code, one can see that this file is created if the usual one cannot be created at `/run/user/1000`:
https://github.com/davatorium/rofi/blob/next/source/rofi.c#L892
Thus, two solutions arise:
1. just provide them with the workaround
2. teach people how to fix their problems with `/run/user/1000`

My first commit titled "Proposed init system agnostic fix" tries to do the first, as the problem relates to what init system one uses. This, obviously, brings maintainability concerns with it:
- If we support x init system's solution, will we have to also include y, z and t init system? Do we have a list of init systems we provide solutions for (the ones we support)? 
- Should a piece of software concerning xdg specification compliance even delve into that?
The second commit attemps to provide the solution for OpenRC, and this probably wouldn't happen on systemd, unless one messes up a config at a very deep level.

Syntax obviously open to change, I just did it the way I thought it looked readable, seeing as the project does not seem to enforce guidelines on these.

PR #159, opened 2022-08-08
update azure-cli to conform with other program names

andrewmustea:
- updated name from python-azure-cliAU.json to azure-cli.json


b3nj5m1n:
Thank you!

PR #161, opened 2022-08-11
add an unsupported status for gitmoji-cli

ccoVeille:
Here is the tool https://github.com/carloscuesta/gitmoji-cli

and the current status https://github.com/carloscuesta/gitmoji-cli/issues/893


b3nj5m1n:
Thanks!

PR #162, opened 2022-08-13
Add support for powerlevel10k

coloursofnoise:
It may also be worth adding a note about consolidating the p10k cache files if that's within scope for this project: https://github.com/romkatv/powerlevel10k/issues/1817


b3nj5m1n:
Thanks, since the cache files are all in the xdg cache directory, I think we'll leave it as it is for now.

PR #175, opened 2022-08-31
add support for vscode-server

andrewmustea:
- Each of vscode's remote development extensions will create and use a
  `vscode-server` directory in a user's home folder each time a host is
  connected to.
- Unfortunately, the SSH extension is the only extension that has a
  setting that can change the location of `vscode-server`. Both the
  container extension and the WSL extension are missing the option of
  changing their `vscode-server` directories.
- Even so, the SSH extension only allows you to define paths for
  individual hosts. It doesn't allow for a default path for all ssh
  connections.
- To edit the host paths, you either need to change your remote SSH
  settings in the vscode application itself, or edit your vscode user
  settings json file.


b3nj5m1n:
Thank you!

PR #181, opened 2022-09-25
Add OmniSharp

m-lamonaca:
From https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Host/Services/OmniSharpEnvironment.cs


b3nj5m1n:
Thanks!

PR #189, opened 2022-11-09
Fixed correct location of exporting ZDOTDIR

realitymolder:
"You can do this in _/etc/zsh/zshenv_ " have been replaced by "You can do this in _/etc/zshenv_" as this is the correct location. Was fighting with this one for a while until I realized the correct location.

### DIFF
--- a/programs/zsh.json
+++ b/programs/zsh.json
@@ -1,49 +1,49 @@
 {
     "files": [
         {
-            "path": "$HOME/.zcompcache",
+            "help": "Set this in your zshrc:\n\n```bash\nzstyle ':completion:*' cache-path \"$XDG_CACHE_HOME\"/zsh/zcompcache\n```\n",
             "movable": true,
-            "help": "Set this in your zshrc:\n\n```bash\nzstyle ':completion:*' cache-path \"$XDG_CACHE_HOME\"/zsh/zcompcache\n```\n"
+            "path": "$HOME/.zcompcache"
         },
         {
-            "path": "$HOME/.zcompdump",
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
Hi, first of all, thanks for bringing my attention to this.

I've actually had a ton of trouble getting this to work in the last 3 days while setting up a new system, and it's this exact issue.

Here's the thing, it seems like it can be either, accoring to the [arch wiki](https://wiki.archlinux.org/title/Zsh#Startup/Shutdown_files), the way it's currently explained by xdg-ninja is correct, and that way works on my arch install. On NixOS, your solution is the only one that works.

I can't look into it right now, but it seems like this is distro specific for some reason? So unless we find more info we probably have to add a note that it could be either.

realitymolder:
> Hi, first of all, thanks for bringing my attention to this.
> 
> I've actually had a ton of trouble getting this to work in the last 3 days while setting up a new system, and it's this exact issue.
> 
> Here's the thing, it seems like it can be either, accoring to the [arch wiki](https://wiki.archlinux.org/title/Zsh#Startup/Shutdown_files), the way it's currently explained by xdg-ninja is correct, and that way works on my arch install. On NixOS, your solution is the only one that works.
> 
> I can't look into it right now, but it seems like this is distro specific for some reason? So unless we find more info we probably have to add a note that it could be either.

Very interesting. According to my testing, this was the issue on macOS + Fedora. I cannot guarantee on other Linux distros. But have you seen this only on Arch wiki? Maybe they are the one that made a mistake in the first place? :O

b3nj5m1n:
It seems `/etc/zshenv` is the offial location, so I added a note that it might be different on some systems.

I also made the files a bit more consistent with each other.

Let me know if you're happy with these changes and I'll merge this!

realitymolder:
Yes! Looks good to me ! Merge away, thanks ! :)

PR #193, opened 2022-11-23
Add .acme.sh

JacobTravers:
https://github.com/acmesh-official/acme.sh/issues/4408

### DIFF
--- a/programs/acme-sh.json
+++ b/programs/acme-sh.json
@@ -1,0 +1,10 @@
+{
+    "files": [
+        {
+            "path": "$HOME/.acme.sh",
+            "movable": false,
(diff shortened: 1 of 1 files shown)


programs/acme-sh.json
```diff
@@ -0,0 +1,10 @@
+{
+    "files": [
+        {
+            "path": "$HOME/.acme.sh",
+            "movable": false,
```
b3nj5m1n, line 5:
```suggestion
            "movable": true,
```
Hi, first of all, thanks for your contribution.

If I understand correctly, the files can be moved, so movable should be set to true.

PR #195, opened 2022-11-23
Add cinnamon and add json schema for program.json

### LINKED ISSUE #194: `.cinnamon` (and `.themes` and `.icons`?)
This is a reminder for myself when I have the time to open a PR, but anyone can open it.

## From Cinnamon version 5.6.0
These are backwards compatible (read from new and old dirs in this order, write to new dirs unless it's modifying an existing file). Moving them is the best action to take.
`.cinnamon/panel-launchers/` -> `$XDG_DATA_HOME/cinnamon/panel-launchers/`
`.cinnamon/backgrounds/` -> `$XDG_DATA_HOME/cinnamon/backgrounds/`
`.cinnamon/configs/` -> `$XDG_CONFIG_HOME/cinnamon/spices/`

Bit unclear if anything else uses them as well, but if it's only Cinnamon then they can be safely moved to the new location. Cinnamon used them to store downloaded themes (and their icon/cursor themes).
`.themes/` -> `$XGD_DATA_HOME/themes/`
`.icons/` -> `$XDG_DATA_HOME/icons/`

These are recreated at the new paths. The ones in the old path can be safely deleted.
`.cinnamon/glass.log` -> `$XDG_STATE_HOME/cinnamon/glass.log`
`.cinnamon/harvester.log` -> `$XDG_STATE_HOME/cinnamon/harvester.log`

Cache, old path can be safely deleted.
`.cinnamon/spices.cache/` -> `$XDG_CACHE_HOME/cinnamon/spices/`

## Commits refs

https://github.com/linuxmint/cinnamon/commit/74c26ab74e7ae0a46c134c62ac2e537d4ec427e6
https://github.com/linuxmint/cinnamon/commit/c0986743cad1f3daa39425b5a125b22d7b6f1daa
https://github.com/linuxmint/cinnamon/commit/0fc32e2047ec9e451034ebd87a50caeb240eb6c4
https://github.com/linuxmint/cinnamon/commit/323b5aa3f3070168956c46cbb0ba140fe66215e8
https://github.com/linuxmint/cinnamon/commit/844493a8279bb326c4ff79490d8f39c0d539960d
https://github.com/linuxmint/cinnamon/commit/f3729147cb652b2e13969a76fb861a276e602ee3

Gr3q:
The schema can help to validate `.json` files in the `programs` folder (either by using a validator or just using it in an editor that supports it). Helpful for people who just wants to write the json file directly and not use the tool.

Also fixes #194 

PS.: I know I should have split this to 2 PRs but it's late and I'm lazy (also it's not a lot)

### DIFF
--- a/program-schema/file.json
+++ b/program-schema/file.json
@@ -1,0 +1,24 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "File schema",
+    "properties": {
+        "path": {
+            "type": "string",
+            "description": "Path to file (or folder).",
+            "minLength": 1
+        },
+        "movable": {
+            "type": "boolean",
+            "description": "Is file (or folder) movable to a place other than $HOME."
+        },
+        "help": {
+            "type": "string",
+            "description": "Help text for user. Supports markdown."
+        }
+    },
+    "required": [
+        "path",
+        "movable",
+        "help"
+    ]
+}
--- a/program-schema/program.json
+++ b/program-schema/program.json
@@ -1,0 +1,24 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "Program",
+    "description": "Specification of files or folders in the $HOME folder for a program",
+    "properties": {
+        "name": {
+            "type": "string",
+            "description": "Name of the program",
+            "minLength": 1
+        },
+        "files": {
+            "description": "List of files (or folders) associated with the program",
+            "type": "array",
+            "items": {
+                "$ref": "file.json"
+            },
+            "minItems": 1
+        }
+    },
+    "required": [
+        "name",
+        "files"
+    ]
+}
--- a/programs/cinnamon.json
+++ b/programs/cinnamon.json
@@ -1,0 +1,46 @@
+{
+    "$schema": "../program-schema/program.json",
+    "name": "cinnamon",
+    "files": [
+        {
+            "path": "$HOME/.cinnamon/spices.cache/",
+            "movable": true,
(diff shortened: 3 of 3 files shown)


b3nj5m1n:
Hi! First of all, thanks for your contribution.

I'm pretty much on board with all the cinnamon stuff, except for `.themes` and `.icons`, you already mentioned this in your original issue, cinnamon is definitely not the only one using these directories. It makes more sense to add a note at the bottom of the existing records for these directories about cinnamon, rather than putting them in the cinnamon file.

I'm slightly conflicted about the whole schema validation thing, mainly because it seems like we'd have to go and add the `$schema` field to every existing config, which I don't want to do.

In general, I think this would be a good idea, especially if we can integrate it with CI so I don't always have to manually check the PRs to see if the JSON is correct. I couldn't immediately find anything on GitHub marketplace when searching for "JSON schema" or "JSON validator", even though I'm sure there is something somewhere.

I'd really like to get rid of the `$schema` field, it shouldn't be a big deal since all the JSON files have to conform to the same schema. I'm not at all familiar with JSON schema, though. I'd also like to rename the directory from `program-schema` to `json-schema`, I think that makes it a bit clearer.

It'd be really cool if you could split this into 2-3 PRs after all.

1. With the cinnamon stuff, minus `.themes` and `.icons` and the `$schema` field in cinnamon.json
2. If you want, an extra PR for `.themes` and `.icons`, or I can add that if you want
3. One for the JSON schema files

Otherwise, I can merge in the cinnamon stuff myself in the next few days and, after I've familiarised myself with JSON schema and the tooling around it a bit more, those files as well. I don't know when I'll get to it, though.

Again, thanks for your contribution!

PR #197, opened 2022-11-24
Add json schema for json validation (for manual editing)

Gr3q:
I know you can omit `$schema` props in VS Code if you have this in your settings:

```json
"json.schemas": [
        {
            "fileMatch": [
                "programs/*.json"
            ],
            "url": "./json-schema/program.json"
        }
    ]
```

Also this looks something that seems more or less fresh (currently) and could be used: https://github.com/marketplace/actions/validate-json-schemas

I know other languages has libraries that can validate against a json-schema, but I couldn't find one that is actively maintained for Haskell.


b3nj5m1n:
> Also this looks something that seems more or less fresh (currently) and could be used: github.com/marketplace/actions/validate-json-schemas

Thanks, I don't know how I managed to miss that. I'll look into setting that up as soon as I have the time.

PR #199, opened 2022-11-25
Updated support of audacity

SpaghettiBorgar:
Audacity follows XDG spec since version 3.2.0 unless the folder already exists in the old location.
See audacity/audacity#3088


b3nj5m1n:
Very nice, thank you!

PR #200, opened 2022-12-02
Add `imwheel`

jackbrower:


b3nj5m1n:
Hi, first of all, thanks for your contribution!

I'm a bit confused by your description, since only `/etc/X11/imwheel/imwheelrc` and `$HOME/.imwheelrc` are considered, symlinking to `$XDG_CONFIG_HOME/imwheelrc` isn't going to do anything.

I'll suggest some changes based on what I gathered from the manpage.

programs/imwheel.json
```diff
@@ -0,0 +1,10 @@
+{
+    "name": "imwheel",
+    "files": [
+        {
+            "path": "$HOME/.imwheelrc",
+            "movable": false,
+            "help": "imwheel does not does not follow XDG.\n\nNOTE: imwheel has a global config. Check the imwheel manpage to see where your system stores this.\n\n```bash:\nsudo ln -s $XDG_CONFIG_HOME/imwheelrc <location of global config>\n```\n"
```
b3nj5m1n, line 7:
```suggestion
            "help": "imwheel does not does not support changing the config file location, but you can configure it through a global config file, probably stored at `/etc/X11/imwheel/imwheelrc`.\n```\n" 
```

jackbrower:
> Hi, first of all, thanks for your contribution!
> 
> I'm a bit confused by your description, since only `/etc/X11/imwheel/imwheelrc` and `$HOME/.imwheelrc` are considered, symlinking to `$XDG_CONFIG_HOME/imwheelrc` isn't going to do anything.
> 
> I'll suggest some changes based on what I gathered from the manpage.

Hi, I agree what I wrote could have been phrased a little better. The intent was to show a workaround where you could write a config file in your home and then link it to the global config (which I found may differ based on distro or version). I tested this locally and it worked as intended.

b3nj5m1n:
Usually, symlinking from a local to global config file is not something you want to do, and I don't think we should suggest it. I'll apply my suggestion and merge this for now, if there's any specific reason you'd want to do it the way you suggested, feel free to open an issue.

PR #203, opened 2022-12-08
Add wakatime

realitymolder:
According to the arch wiki doc : https://wiki.archlinux.org/title/XDG_Base_Directory

image


b3nj5m1n:
Thanks!

PR #209, opened 2023-01-10
Add configs for various programs

ShivanshuKantPrasad:
Configs for nix, renderdock, tldr, Networkmanager-openvpn and xdg-utils.

Networkmanagegr-openvpn, xdg-utils and renderdock info is from arch wiki.


b3nj5m1n:
Thank you!

PR #213, opened 2023-01-13
Add lens

NyCodeGHG:
Lens seems to use some xdg directories for parts of it, but not for its extensions directory.
Relevant Issue: https://github.com/lensapp/lens/issues/2494


b3nj5m1n:
Thanks!

PR #215, opened 2023-01-25
`~/.gitignore` can be moved to `XDG_CONFIG_HOME/git/ignore`

rvenutolo:
https://git-scm.com/docs/gitignore indicates it will read ignore configuration from `$XDG_CONFIG_HOME/git/ignore`.


b3nj5m1n:
Thanks!

PR #216, opened 2023-01-25
Add `nano`

rvenutolo:
See https://savannah.gnu.org/patch/?8523 and https://git.savannah.gnu.org/cgit/nano.git/commit/?id=c16e79b


b3nj5m1n:
Thank you!

PR #217, opened 2023-01-28
Added option to specify the source of the information displayed

LukasPietzschmann:
The source entry could contain a link that points the user in the right direction if he/she wants to research the given help by him/herself.

## Todo:
- [ ] Implement this feature into `xdgnj` so it can be automatically generated (I'm pretty new to Haskell, so this may take some time until I understood the code good enough to implement this)

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,8 @@ You define a program, and then a list of files and directories which this progra
 
 For each file/directory, you specify if it can be (re)moved.
 
-If this is the case, you also specify instructions on how to accomplish this in Markdown.
+If this is the case, you also specify instructions on how to accomplish this in Markdown. Alongside the instruction you can also
+specify a source to let the user know where you got this information from.
 
 Files in this directory can have any name, but using the name of the program is encouraged.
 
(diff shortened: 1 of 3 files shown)


b3nj5m1n:
I very much agree that it would be useful to include a link to the source, but I'm not sure if making it a separate field is the right way to go.

Adding a source link to all existing help pages is unrealistic, so this would have to be optional in any case, and at that point I think it would be sufficient to include a link in the help itself.

We already try to encourage linking to an issue page or something like that for programs which do not support xdg. 

https://github.com/b3nj5m1n/xdg-ninja/blob/897ce07a19e33c78105449898ae25c6049467aad/lib/AddProgram.hs#L30

This system seems to have worked pretty well so far, we could also put something like that in the templates for supported programs, which I think is probably better than creating a dedicated field for it.

If you disagree, let me know if there's any specific reason for why you think making it a dedicated field would be more beneficial.

LukasPietzschmann:
I thought a dedicated field could be nice as there would be a central point where such links are specified. So we could “control” how and where it is displayed, and the user would not have to scan through the help text to search for it.

But I totally see your point that it might be unnecessary to add a whole new field if there's already a system that's working just fine.

PR #218, opened 2023-02-01
`~/.fonts.conf` can be moved to `$XDG_CONFIG_HOME/fontconfig/fonts.conf`

ChaseCares:
https://www.freedesktop.org/software/fontconfig/fontconfig-user.html states that `~/.fonts.conf` can be placed in several different locations, most notably `$XDG_CONFIG_HOME/fontconfig/fonts.conf`. This pull request adds that recommendation.

Thanks


b3nj5m1n:
Thanks!

PR #219, opened 2023-02-10
Add `hammerspoon`

swt30:
Hammerspoon is a program for scripting MacOS; it puts config
in `~/.hammerspoon` by default. The location can be changed with the
MacOS `defaults` command.

See https://github.com/Hammerspoon/hammerspoon/issues/2175


b3nj5m1n:
Thank you!

PR #221, opened 2023-02-15
keychain: add `--absolute` opt to avoid subdirectory

lediur:
## Description

This PR adds `--absolute` to the command provided in the keychain migration instructions.

## Motivation

For backwards compatibility, if the user provides a path `--dir` to keychain, it will create a `.keychain` folder in that path. Adding the [`--absolute`](https://man.archlinux.org/man/keychain.1#absolute) opt will encourage keychain to use the directory as is.

### DIFF
--- a/programs/keychain.json
+++ b/programs/keychain.json
@@ -3,7 +3,7 @@
         {
             "path": "$HOME/.keychain",
             "movable": true,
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
Thank you!

PR #226, opened 2023-03-12
Add flake.nix

nullishamy:
This adds support for Nix flakes.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,32 +1,37 @@
 {
-  inputs.flake-utils.url = "github:numtide/flake-utils";
+  inputs = {
+    nixpkgs.url = "nixpkgs";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
   outputs = {
     self,
     nixpkgs,
     flake-utils,
   }:
-    flake-utils.lib.eachDefaultSystem (
-      system: let
-        pkgs = nixpkgs.legacyPackages.${system};
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        runtimeDependencies = with pkgs; [
+          glow
+          jq
+        ];
+        overlays = [
+          (self: super: {
+            xdg-ninja = super.xdg-ninja.overrideAttrs (old: {
+              version = "git";
+              src = ./.;
+            });
+          })
+        ];
+        pkgs = import nixpkgs { inherit system overlays; };
       in rec {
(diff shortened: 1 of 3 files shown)


b3nj5m1n:
Hi! First of all, thank you for this!

I'm currently in the process of nix'ifying a lot of my personal projects, so great timing as well.

I have to admit that I'm a little confused about where this package definition is supposed to go, since it's already a [package](https://github.com/NixOS/nixpkgs/blob/nixos-22.11/pkgs/tools/misc/xdg-ninja/default.nix) in the official nixpkgs repo.

It doesn't seem like a good idea to maintain both, but it also seems like a good idea to maintain it as part of a flake in the project's repository.

I guess my question is what the current best practice is regarding this, would the nixpkgs definition be changed to use this flake, should we change the definition of this flake to use the nixpkgs definition and overwrite the `src` attribute?

<sub>I'd also like to add a dev shell, but I can do that myself after this PR is merged.</sub>

flake.nix
```diff
@@ -0,0 +1,32 @@
...
+          xdg-ninja = pkgs.stdenvNoCC.mkDerivation {
+            name = "xdg-ninja";
+            src = ./.;
+            nativeBuildInputs = [pkgs.makeWrapper];
+            installPhase = ''
+              runHook preInstall
+              install -Dm755 xdg-ninja.sh "$out/share/xdg-ninja/xdg-ninja.sh"
+              install -Dm644 programs/* -t "$out/share/xdg-ninja/programs"
+              mkdir -p "$out/bin"
+              ln -s "$out/share/xdg-ninja/xdg-ninja.sh" "$out/bin/xdg-ninja"
```
SuperSandro2000, line 22: Why not move it to $out/bin in the first place?

nullishamy:
Oh, I did not see the official package :laughing:

I'm not quite sure what the etiquette there is. Maybe opening something on the Nix Discourse would be a good idea?

flake.nix
nullishamy: `install` handles file permissions (setting the executable flag) at the same time.

b3nj5m1n:
Since I'm a little busy at the moment, for now I just did what seemed the most sensible to me. (Use the derivation <sup>I hope I'm using that term correctly</sup> from the official nixpkgs repo and overwrite the `src` attribute)

Thank you again for the contribution!

PR #228, opened 2023-03-16
Add `rusherhack` support

REALERvolker1:
Rusherhack is a Minecraft anarchy mod. It puts the auth files in `~/.rusherhack` by default. This will likely [change](https://github.com/RusherDevelopment/rusherhack-issues/issues/1422) in the future, but for now, this is warranted.


b3nj5m1n:
Thank you!

PR #229, opened 2023-03-20
add `octave` to darabase

CheesyChocolate:
* octave fully supports XDG for history file and site packages.
* octave partially supports XDG for rc file. An environment variable is needed.


b3nj5m1n:
Thanks!

PR #231, opened 2023-03-28
Add fzf

mxmlnkn:
This also affects `~/.fzf.bash` and possibly `~/.fzf.zsh` but those are removed by the uninstall script. I'm unsure whether to list both files separately in the json file.

### DIFF
--- a/programs/fzf.json
+++ b/programs/fzf.json
@@ -1,0 +1,10 @@
+{
+    "files": [
+        {
+            "path": "$HOME/.fzf",
+            "movable": true,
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
> This also affects ~/.fzf.bash and possibly ~/.fzf.zsh

In theory, we should add them. (Although there are probably even more for other shells)

Is XDG respected by default in newer versions? Since it seems like the minimum version you specified has been out for quite a while, I don't think it's necessary to add all individual files if new users will never see them.

If the default behavior of fzf is still to create these files in the home directory, though, I think we should add all of the files it might create.

I'll go ahead and merge this now, I might have time to look into this myself today.

mxmlnkn:
> Is XDG respected by default in newer versions?

When installing from source, no. The `--xdg` option is still necessary. I don't know how it behaves when installing it via the package manager. I know that in that case there are some manual instructions for installing the bash/zsh bindings. I assume that those also install `.fzf.bash` but not the `.fzf` folder, which contains the git repository and the compiled binary among others.

PR #236, opened 2023-03-31
Add kotlin konan

zt64:
This just adds the `.konan` folder. I'm unsure if there's any other kotlin related folders/files in the home dir so I just named it as `kotlin` in case of future additions


b3nj5m1n:
Thank you!

PR #238, opened 2023-04-01
Add Aegisub

Zweihander-Main:
Aegisub is a subtitle editor program. Creates `$HOME/.aegisub`. Relevant issue referenced.


b3nj5m1n:
And thanks again!

PR #239, opened 2023-04-01
Add EqualX

Zweihander-Main:
EqualX is a GUI LaTeX editor. It creates `$HOME/.equalx`. Relevant issue referenced.


b3nj5m1n:
Thanks!

PR #245, opened 2023-04-02
Add electron builders

remgodow:
Adds information about directories created by various electron apps (only when building them from source).


b3nj5m1n:
Thank you!

PR #247, opened 2023-04-07
NNormalize referring to environment variables for max compatibility.

rpavlik:
- Remove curly brackets for better copy/paste compat with fish
- Add $ to mentions of variables in paths that do not already have them

### DIFF
--- a/programs/alsa.json
+++ b/programs/alsa.json
@@ -4,7 +4,7 @@
         {
             "path": "$HOME/.asoundrc",
             "movable": true,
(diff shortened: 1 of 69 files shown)


b3nj5m1n:
Thank you very much for this!

May I ask how you did this? If you wrote a script or something, maybe we could convert that to a GitHub action to test if new PRs comply with this automatically?

rpavlik:
Just a bunch of search and replace in tree. We could probably do a sed script though.

PR #248, opened 2023-04-08
Add mamba

erooke:
[Mamba](https://mamba.readthedocs.io/en/latest/index.html) currently does not support XDG, added information to point to the [issue](https://github.com/mamba-org/mamba/issues/1787)


b3nj5m1n:
Thank you!

Issue #204, opened 2022-12-08
Homebrew version is outdated 


realitymolder:
Hey, seems like the brew version of the program isn't updated, at least since May. (maybe even longer)

b3nj5m1n:
I'm not really familiar with brew, but it seems like the package is pinned to the v0.2.0.1 tag, which is unfortunate since the tags are really meant more for the haskell program than the script.

If you know more about homebrew and could help with this, here is the PR which originally added xdg-ninja to homebrew: https://github.com/Homebrew/homebrew-core/pull/104762.

realitymolder:
> I'm not really familiar with brew, but it seems like the package is pinned to the v0.2.0.1 tag, which is unfortunate since the tags are really meant more for the haskell program than the script.
> 
> If you know more about homebrew and could help with this, here is the PR which originally added xdg-ninja to homebrew: [Homebrew/homebrew-core#104762](https://github.com/Homebrew/homebrew-core/pull/104762).

Ok, so I did some research, it seems like the progress is manual. 
It is mostly dependent on releasing a new binary, but I see that it hasn't been made since May. 
So I will suggest 2 solutions:
1. Release a binary in the repo "release page" every month, a new version with the newest binaries. This will then be updated manually in the brew formula.
2. Add a function to update the script / JSON database from the script itself - after it has been made, update the Homebrew version "for the last time".

patricksjackson:
The version in [`nixpkgs`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/misc/xdg-ninja/default.nix) also uses the release tags fyi.

These package managers will always be pinned to _something_, so when there are releases that's generally what they will use. From the package manager's perspective it is a lot easier if a project has occasional releases but they are usually able to just pin to random git commits if necessary.

b3nj5m1n:
> but they are usually able to just pin to random git commits if necessary.

At least for nix that's the case, and I think that's probably the best way to go in this case, since it doesn't really make sense to create a release for every program added, and I'm not sure what else to base it on, except for maybe periodical releases every few months.

KaBankz:
You can always install using the `--HEAD` flag to install from the latest commit.

```sh
brew install xdg-ninja --HEAD
```

PR #252, opened 2023-04-22
Unified packaging process

Ballasi:
Hey!

It seems to me that xdg-ninja is packaged quite poorly in some cases, like [putting a symlink to the programs/ directory directly in the bin/ folder of the system](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=xdg-ninja) or [modifying the script in place to patch the location of the programs/ directory](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=xdg-ninja-git).

This mostly is a consequence of the way the script currently fetches the `programs/` directory:

https://github.com/b3nj5m1n/xdg-ninja/blob/4644d409aaf00b7d3b6864840f23e4c05c830322/xdg-ninja.sh#L221

i.e., directly in the current working directory of the xdg-ninja script.

This works fine when working within the git repository but does poorly when trying to package the program as there needs to implement hacks to get this working.

_On a useless note, I wouldn't have hacked it this way on my side, I would've just packaged xdg-ninja in the same directory as programs in, for instance, `/usr/share/xdg-ninja/` and would create a wrapper script that would call `/usr/share/xdg-ninja/xdg-ninja.sh`._

I am unsure whether or not this is a goal of your program to be packaged, but if so, I have implemented an simple way to unify the packaging process.

---

My main goal, when modifying the script, was to get it backwards compatible with how it was previously used, i.e. when working in the git repository, it does fetch the current `programs/` dir rather than somewhere else.

So I have just set the script to check in the current working directory and replace ending instances of `/bin` to `/share/xdg-ninja` (where the `programs/` dir is located when packaged).

If all cases it fails, the `XN_PROGRAMS_DIR` can be used to override that value. I have preferred to use this instead of a `--programs-dir=` argument as this isn't something meant for common usage either.

I've also allowed people to override where the script is supposed to be located with the `PREFIX` variable in the `Makefile`, I believe package managers should package apps to `/usr` instead of `/usr/local` so at least it can be changed. This `make` variable and `DESTDIR` and both common naming standards.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,18 @@
+PREFIX ?= /usr/local
+
+all:
+	@echo "Usage: make (install|uninstall)"
+
+.PHONY: all install uninstall
+
+install:
+	install -Dm 0755 xdg-ninja.sh $(DESTDIR)$(PREFIX)/bin/xdg-ninja
+	install -d $(DESTDIR)$(PREFIX)/share/xdg-ninja/
+	cp -r programs $(DESTDIR)$(PREFIX)/share/xdg-ninja/
+	install -d $(DESTDIR)$(PREFIX)/share/doc/xdg-ninja/
+	install -m 0644 LICENSE README.md $(DESTDIR)$(PREFIX)/share/doc/xdg-ninja/
+
+uninstall:
+	rm -rf $(DESTDIR)$(PREFIX)/bin/xdg-ninja \
+	       $(DESTDIR)$(PREFIX)/share/xdg-ninja \
+	       $(DESTDIR)$(PREFIX)/share/doc/xdg-ninja
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ To install xdg-ninja with [Homebrew](https://brew.sh), run `brew install xdg-nin
 
 ## Configuration
 
-The configuration is done in the _programs/_ directory.
+The configuration is done in the _programs/_ directory, which should be located in the same working directory as the xdg-ninja.sh script. This can be overriden with the `XN_PROGRAMS_DIR` environment variable.
 
 You define a program, and then a list of files and directories which this program ruthlessly puts into your _$HOME_ directory.
 
--- a/flake.nix
+++ b/flake.nix
@@ -26,7 +26,25 @@
       in rec {
         packages = flake-utils.lib.flattenTree {
           # The shell script and configurations, uses derivation from offical nixpkgs
-          xdg-ninja = pkgs.xdg-ninja;
+          xdg-ninja = pkgs.stdenv.mkDerivation rec {
+            pname = "xdg-ninja";
+            version = "0.1.0";
+
+            src = ./.;
+
+            nativeBuildInputs = with pkgs; [ makeWrapper ];
+
+            installPhase = ''
+              runHook preInstall
+
+              DESTDIR="$out" PREFIX="/usr" make install
+
+              wrapProgram "$out/usr/bin/xdg-ninja" \
+                --prefix PATH : "${pkgs.lib.makeBinPath [ pkgs.glow pkgs.jq ]}"
+
+              runHook postInstall
+            '';
+          };
           # Pre-built binary of xdgnj tool downloaded from github
           xdgnj-bin = pkgs.stdenvNoCC.mkDerivation {
             name = "xdgnj-bin";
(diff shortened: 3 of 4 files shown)


b3nj5m1n:
Hi! Thank you very much for this, it seems like a much-needed change.

The only concern I have is with the `Install` section in the README. I'm not sure that it makes sense to present this to users as the best/only option.

There is another [big issue](https://github.com/b3nj5m1n/xdg-ninja/issues/204) with packaging, I think perhaps I should finally think of something to fix that, update the packages, (I'm aware of arch, homebrew and nix) then write a proper installation section which recommends installing through a package manager first. The README needs a bit of an overhaul anyway tbh.

If you're okay with it, I'd say remove the `Install` section for now.

And again, thanks for putting time into this!

---

<sup>Btw, this breaks the nix flake, I tried to update it but it seems I don't have write permissions for this PR, I'll just leave the patch here, if you want to apply it that'd be great, otherwise I'll do it after merging.</sup>

<details>

```patch
From 640977d61030c576da3b47edfa09fd651b75a6ee Mon Sep 17 00:00:00 2001
From: b3nj5m1n <b3nj4m1n@gmx.net>
Date: Sun, 23 Apr 2023 08:00:13 +0200
Subject: [PATCH] Update flake

---
 flake.nix | 22 ++++++++++++++++++++--
 1 file changed, 20 insertions(+), 2 deletions(-)

diff --git a/flake.nix b/flake.nix
index 91adf01..8f3a5bf 100644
--- a/flake.nix
+++ b/flake.nix
@@ -26,7 +26,25 @@
       in rec {
         packages = flake-utils.lib.flattenTree {
           # The shell script and configurations, uses derivation from offical nixpkgs
-          xdg-ninja = pkgs.xdg-ninja;
+          xdg-ninja = pkgs.stdenv.mkDerivation rec {
+            pname = "xdg-ninja";
+            version = "0.1.0";
+
+            src = ./.;
+
+            nativeBuildInputs = with pkgs; [ makeWrapper ];
+
+            installPhase = ''
+              runHook preInstall
+
+              DESTDIR="$out" PREFIX="/usr" make install
+
+              wrapProgram "$out/usr/bin/xdg-ninja" \
+                --prefix PATH : "${pkgs.lib.makeBinPath [ pkgs.glow pkgs.jq ]}"
+
+              runHook postInstall
+            '';
+          };
           # Pre-built binary of xdgnj tool downloaded from github
           xdgnj-bin = pkgs.stdenvNoCC.mkDerivation {
             name = "xdgnj-bin";
@@ -45,7 +63,7 @@
         };
         defaultPackage = packages.xdg-ninja;
         apps = {
-          xdg-ninja = flake-utils.lib.mkApp { drv = packages.xdg-ninja; };
+          xdg-ninja = flake-utils.lib.mkApp { drv = packages.xdg-ninja; exePath = "/usr/bin/xdg-ninja"; };
           xdgnj-bin = flake-utils.lib.mkApp { drv = packages.xdgnj-bin; exePath = "/bin/xdgnj"; };
         };
         defaultApp = apps.xdg-ninja;
-- 
2.39.2
```

</details>

Ballasi:
I haven't taken the time to take a look at nix as I am unfamiliar with this distro, so that's a good thing you were able to test that on your side. I have tested this on Ubuntu and Arch.

I have added your patch on my branch and removed the Install section of the README as discussed.

b3nj5m1n:
Great, thank you very much again!

Ballasi:
No problem, I can see that the AUR package for `xdg-ninja-git` has [already been updated](https://aur.archlinux.org/cgit/aur.git/commit/?h=xdg-ninja-git&id=4f5c6d69646ac9711c928983a01ebce1fde7d78e) too!

From there, I can only recommend you to do a new GitHub release to push an update to the `xdg-ninja` AUR package which relies on tags/releases rather than the git repository version itself.

b3nj5m1n:
> I can only recommend you to do a new GitHub release

Yeah, I'm still thinking about how to handle releases, I could either use them exclusively for xdg-ninja, (the script & new json files) or have a prefix for xdg-ninja/xdgnj to distinguish between the two.

> I can see that the AUR package for xdg-ninja-git has [already been updated](https://aur.archlinux.org/cgit/aur.git/commit/?h=xdg-ninja-git&id=4f5c6d69646ac9711c928983a01ebce1fde7d78e) too!

Wow that was quick!

PR #257, opened 2023-05-06
Fix npmrc env var format

mattico:
npmrc only supports interpolating env vars with braces like this.

https://docs.npmjs.com/cli/v8/configuring-npm/npmrc#files


b3nj5m1n:
Thanks!

PR #258, opened 2023-05-10
man: create xdg-ninja.1 man page

Ballasi:
Because man pages are cool.

You can preview it by using `$ man man/xdg-ninja.1`

I can also edit the man page if you think it might require some changes before pushing. I was wondering about adding a paragraph for the optional dependencies (glow and alternatives) but figured it would be a better fit to have that in the `README.md` as a one-time installation information.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,8 +11,10 @@ install:
 	cp -r programs $(DESTDIR)$(PREFIX)/share/xdg-ninja/
 	install -d $(DESTDIR)$(PREFIX)/share/doc/xdg-ninja/
 	install -m 0644 LICENSE README.md $(DESTDIR)$(PREFIX)/share/doc/xdg-ninja/
+	install -m 0644 man/xdg-ninja.1 $(DESTDIR)$(PREFIX)/share/man/man1/
 
 uninstall:
 	rm -rf $(DESTDIR)$(PREFIX)/bin/xdg-ninja \
 	       $(DESTDIR)$(PREFIX)/share/xdg-ninja \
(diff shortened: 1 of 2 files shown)


b3nj5m1n:
Very cool, thank you!

It seems there's a small typo, the options are named as `strip-ok`, `strip-unsupported`, etc., when they're actually named `skip-ok`, `skip-unsupported`.

> I was wondering about adding a paragraph for the optional dependencies

I think there should at least be a paragraph about the recommended dependencies, so `jq` and `glow`, although I'm not really familiar with the best practices for man pages.

Apart from that it looks good!

Ballasi:
> It seems there's a small typo, the options are named as `strip-ok`, `strip-unsupported`, etc., when they're actually named `skip-ok`, `skip-unsupported`.

You are right! I actually misread `strip`, that's my bad. Fixed.

> > I was wondering about adding a paragraph for the optional dependencies
> 
> I think there should at least be a paragraph about the recommended dependencies, so `jq` and `glow`, although I'm not really familiar with the best practices for man pages.

Since `jq` is a required (and not a recommended) dependency and the script fails when running without it, I would tend to not have that in the `man` page, because of the way I see man pages ; documentation on how to use a program and not how to install it.

I think it would make sense to add a note about optional dependencies as the user can have a different behavior depending on that during runtime.

For now I have just added a note about `glow`, if you prefer also adding a note about `jq` too I can squash that in.

b3nj5m1n:
Looks good to me. Thanks again for all your help with this project!

PR #261, opened 2023-05-14
VS Code Insiders build support

e-khurmamatov:
Added support for `.vscode-insiders` directory. It is same as `.vscode` but used by insiders build.


programs/vscode.json
```diff
@@ -5,6 +5,11 @@
             "movable": false,
             "help": "Currently unsupported.\n\n_Relevant issue:_ https://github.com/microsoft/vscode/issues/3884 \n"
         },
+        {
```
e-khurmamatov, line 8: Thought that there is no need in additional file for VS Code Insiders, so moved it into `vscode.json`. 

Commit 009610a70647f6aa4ffa266227e1159eb8fa6b16 can be reverted to use `vscode-insiders.json`.

b3nj5m1n:
Thanks!

Issue #263, opened 2023-05-16
Improve path matching


Midblyte:
A few suggestions about the JSON syntax, to make the file path more dynamic.

- **"Path" as a list of strings**
Sometimes, two (or more) equal entries were added when it was needed to match two (or more) different paths, even if the entries had the exactly same "help" and "movable" keys (e.g.: [here](https://github.com/b3nj5m1n/xdg-ninja/blob/b11bd6a52c09c1242c1f9a47c010601413e23962/programs/zsh.json) the "help" key was repeated three times for .zhistory, .histfile and .zsh_history).
This is a problem because it may be necessary to edit one message and at the same time forget to edit the others.
Depending on the implementation, it could be necessary to edit every existing .json files to support the new syntax.

- **"Path" as a regex**
Regex patterns allow to match even very different and complex strings.
While I don't personally like the idea (slightly worse performance, uglier paths, more complexity, almost each .json should be fixed because the dot '.' is the match-all character while '\.' is the plain dot, increased risk of matching wrong files, etc...), maybe in certain cases it could be the most flexible way (think of optional alphanumerical suffixes; see following examples).

- **"Path" including variables + fallback**
Currently, it is possible to use shell variables using the dollar sign followed by the variable name (e.g.: $USER).
But this feature can be relied upon only for a very, very small subset of variables (e.g.: it's difficult to even get the hostname since $HOST is Zsh-specific, $HOSTNAME is Bash-specific, while shells like Dash don't even have it).
A possible solution is to create a new key, for example "vars", like this:
`"vars": {"XDG_NINJA_HOSTNAME": ["HOSTNAME", "HOST"]}`
This is how it would work: it picks the first one if it exists, else the second, and so on. If none of them exist, the check on the path should be skipped completely.

- **"Path" including command output + fallback**
Something like the previous paragraph, but for commands. For example:
`"cmds": {"XDG_NINJA_HOSTNAME": ["uname -n", "hostname"]}`.
If none of them has a returncode different from 0, the check on the path should be skipped completely.

_Please note_: I don't know if the added complexity is worth it. To be fair, except for the first suggestion, I doubt it.
[...]

PR #264, opened 2023-05-16
Update zsh.json and ohmyzsh.json

Midblyte:
Add support for `.zshrc.pre-oh-my-zsh`, `.zshrc.pre-oh-my-zsh-YYYY-mm-dd_HH-MM-SS`, `.zcompdump-<hostname>-<version>.zwc`, `.zcompdump-<hostname>-<version>` and `.zcompdump-<user>-<version>`.

~~_Note_: due to current limitations, it's not possible to match a slight variation of this path, namely `.zshrc.pre-oh-my-zsh-YYYY-mm-dd_HH-MM-SS`. See #263.~~

This pull request relies on #271.


b3nj5m1n:
I briefly tested this as part of #271, so I think we're good!

Thank you, and also my apologies to you for taking so long with this.

PR #273, opened 2023-06-06
Update the message for $HOME/R

ArcWandx86:
The existing message for `$HOME/R` was written seemingly with Rstudio in mind. However, the folder is applicable to R as a whole, not just Rstudio (I use nvim-R instead of Rstudio).

The link in the original did not work for me; I found the [specific section of the docs](https://cran.r-project.org/doc/manuals/r-release/R-admin.html#Add_002don-packages) about this from a stackoverflow post.


ArcWandx86:
This will require a `$HOME/.Renviron` file, which isn't quite ideal. I think it's probably possible to put the `.Renviron` file somewhere else, but I have not found out where yet.

b3nj5m1n:
Thanks!

PR #280, opened 2023-06-12
Add calibre

Player01osu:


TinfoilSubmarine:
This seems unneeded. calibre provides the ability to move the library wherever you want (at least as of my distro's packaged version 6.17.0).

b3nj5m1n:
I agree with @TinfoilSubmarine here.

I'm closing this. If there's any specific reason you think this should instead be done through an alias, let me know.

PR #282, opened 2023-06-13
Check for interactive terminal and user request for colorless

Aeyk:
don't initialize color if TERM is dumb, xterm-mono, if NO_COLOR is set, or if LS_COLORS is not 0 (technically this check is for LS_COLORS greater than 0, but i've never seen negative values for it, but if you disapprove of my laziness i can add a greater than -1 test too.)

### DIFF
--- a/xdg-ninja.sh
+++ b/xdg-ninja.sh
@@ -42,7 +42,10 @@ init_constants() {
 
     BG_MAGENTA="\033[45m"
 }
-init_constants
+
+if test -t 0 && test "$TERM" != "dumb" && test "$TERM" != "xterm-mono" && test -n "$NO_COLOR" & test "$LS_COLORS" -gt 0; then
+    init_constants
+fi
 
 help() {
     init_constants


xdg-ninja.sh
```diff
@@ -42,7 +42,10 @@ init_constants() {
 
     BG_MAGENTA="\033[45m"
 }
-init_constants
+
+if test -t 0 && test "$TERM" != "dumb" && test "$TERM" != "xterm-mono" && test -n "$NO_COLOR" & test "$LS_COLORS" -gt 0; then
```
b3nj5m1n, line 46: I believe there's a & missing here
```suggestion
if test -t 0 && test "$TERM" != "dumb" && test "$TERM" != "xterm-mono" && test -n "$NO_COLOR" && test "$LS_COLORS" -gt 0; then
```

b3nj5m1n, line 46: I _think_ this should be `z` instead of `n`?
```suggestion
if test -t 0 && test "$TERM" != "dumb" && test "$TERM" != "xterm-mono" && test -z "$NO_COLOR" & test "$LS_COLORS" -gt 0; then
```
I don't have `$NO_COLOR` set, `test -z "$NO_COLOR"` gives me true, while `test -n "$NO_COLOR"` gives me false.

b3nj5m1n:
> technically this check is for LS_COLORS greater than 0, but i've never seen negative values for it

Here's an example LS_COLORS:
```
➜ echo $LS_COLORS
di=34:ln=35:so=32:pi=33:ex=31:bd=34;46:cd=34;43:su=30;41:sg=30;46:tw=30;42:ow=30;43
```

This makes `test` error:
> test: integer expression expected

I think `LS_COLORS` should probably be ignored completely.

PR #285, opened 2023-06-28
Improve GDB recommendation

dosisod:
The previous suggestion was to use an alias and specify the config path directly, though GDB supports XDG paths according to the man pages.

I don't know what version XDG support was added, or if it has been there all along. I'm currently using GDB version 13.1

### DIFF
--- a/programs/gdb.json
+++ b/programs/gdb.json
@@ -4,7 +4,7 @@
         {
             "path": "$HOME/.gdbinit",
             "movable": true,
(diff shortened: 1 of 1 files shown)


mxmlnkn:
Can confirm with gdb 13.1 on Ubuntu 23.04. Thanks. I didn't find this documented anywhere else. According to the [changelog](https://www.sourceware.org/gdb/news/) it has been added since gdb 11.1, which was released on 2021-09-13.

b3nj5m1n:
Thanks!

PR #299, opened 2023-07-27
updated cups, it is now supported by libcups

karma-riuk:
The relevant issue that was pointed to previously was solved at the beginning of this year, cups does support the XDG spec now.


b3nj5m1n:
Thanks!

PR #316, opened 2023-09-21
Flake cleanup and make usable in a system configuration

SuperSandro2000:
With those changes I can now use the flake directly in my nixos config.


flake.nix
```diff
@@ -37,10 +34,10 @@
             installPhase = ''
               runHook preInstall
 
-              DESTDIR="$out" PREFIX="/usr" make install
+              DESTDIR="$out" PREFIX= make install
 
-              wrapProgram "$out/usr/bin/xdg-ninja" \
-                --prefix PATH : "${pkgs.lib.makeBinPath [ pkgs.glow pkgs.jq ]}"
+              wrapProgram "$out/bin/xdg-ninja" \
+                --prefix PATH : "${pkgs.lib.makeBinPath (with pkgs; [ glow jq ])}"
```
b3nj5m1n, line 40: Because of #271 we also need to include find here:
```suggestion
                --prefix PATH : "${pkgs.lib.makeBinPath (with pkgs; [ glow jq findutils ])}"
```

b3nj5m1n:
Thank you!

PR #318, opened 2023-09-28
Fix skipping one program during check

b3nj5m1n:
When there's only one file in the `programs/` directory, it does not get picked up by xdg-ninja.

Reproducing:
```bash
git clone git@github.com:b3nj5m1n/xdg-ninja.git && cd xdg-ninja
mv programs/abook.json . && rm programs/* && mv abook.json programs
./xdg-ninja.sh -v
```
This doesn't produce any output for abook.json on the main branch.

Since I found this accidentally during testing #271 and "fixed" it using ChatGPT, I'm not sure if this may introduce other unexpected side effects.

### DIFF
--- a/xdg-ninja.sh
+++ b/xdg-ninja.sh
@@ -230,7 +230,7 @@ do_check_programs() {
 " read -r name; read -r filename; read -r movable; read -r help; do
         check_file "$name" "$filename" "$movable" "$help"
     done <<EOF
-$(jq 'if type == "object" then .files[] as $file | .name, $file.path, $file.movable, $file.help else . end' programs/* | sed -e 's/^"//' -e 's/"$//')
+$(jq 'if type == "object" then .files[] as $file | .name, $file.path, $file.movable, $file.help else . end' "$XN_PROGRAMS_DIR"/* | sed -e 's/^"//' -e 's/"$//')
 EOF
 # sed is to trim quotes
 }
--- a/xdg-ninja.sh
+++ b/xdg-ninja.sh
@@ -230,7 +230,7 @@ do_check_programs() {
 " read -r name; read -r filename; read -r movable; read -r help; do
         check_file "$name" "$filename" "$movable" "$help"
     done <<EOF
-$(jq 'inputs as $input | $input.files[] as $file | $input.name, $file.path, $file.movable, $file.help' "$XN_PROGRAMS_DIR"/* | sed -e 's/^"//' -e 's/"$//')
+$(jq 'if type == "object" then .files[] as $file | .name, $file.path, $file.movable, $file.help else . end' programs/* | sed -e 's/^"//' -e 's/"$//')
 EOF
 # sed is to trim quotes
 }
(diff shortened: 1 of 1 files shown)


Ballasi:
From my tests, the bug actually seems to be a bit more critical.

It seems like it skips the first check in **all cases**, not just when there's one file in `programs/`. In the current program, this means skipping the check of FreeCAD.

From what it seems, the issue comes from the `inputs` tag in `jq`, I quickly did a Google search but couldn't find anything in 10s, by replacing it to `.` (what I've often seen in `jq` as being "the root of the JSON file), it works, and it happens to be one of the things ChatGPT did. We would then have something similar to this:

```
jq '. as $input | $input.files[] as $file | $input.name, $file.path, $file.movable, $file.help' "$XN_PROGRAMS_DIR"/*
```

One other thing ChatGPT did is completely get rid of the `$input` variable, since it is now defined as `.` you can just replace the instances of `$input` to virtually nothing (you don't write `$input.name`->`..name` but just `.name`), so you would have something like this:

```
jq '.files[] as $file | .name, $file.path, $file.movable, $file.help' "$XN_PROGRAMS_DIR"/*
```

That would be what I would tend to put in the end.

For some reason, ChatGPT checks if the JSON object is an object (`if type == "object" then`) and spits it out as raw if not (`else . end`). That's probably not a good idea to keep that as it is then given to the `read` command which might mess things up because it might not just return four lines.

It's better to have a catch case in here in case there's an issue (without it, `jq` should just return an error, we need to check that but I believe it makes the app itself crash, so it's already a "catch case" without that change).

I would tent to go with the last line I gave you, not the one ChatGPT gave you. I'll check what's up with that `inputs` thing and what it's used for in `jq`, but that's what causing the issue here.

Ballasi:
Related stackoverflow issue concerning `inputs`: https://stackoverflow.com/a/55996042

xdg-ninja.sh
```diff
@@ -230,7 +230,7 @@ do_check_programs() {
 " read -r name; read -r filename; read -r movable; read -r help; do
         check_file "$name" "$filename" "$movable" "$help"
     done <<EOF
-$(jq 'inputs as $input | $input.files[] as $file | $input.name, $file.path, $file.movable, $file.help' "$XN_PROGRAMS_DIR"/* | sed -e 's/^"//' -e 's/"$//')
+$(jq 'if type == "object" then .files[] as $file | .name, $file.path, $file.movable, $file.help else . end' programs/* | sed -e 's/^"//' -e 's/"$//')
```
Ballasi: you can just use:
```
jq '.files[] as $file | .name, $file.path, $file.movable, $file.help' "$XN_PROGRAMS_DIR"/*
```
no need for the `if type == "object" then [..] else . end` case

b3nj5m1n:
Sorry for taking so long to respond again, I've implemented your suggestion and the issue seems to be fixed, so I think we're good to go.

PR #325, opened 2023-10-22
Add .erlang.cookie

Cantido:
Adds a state file for the [Erlang](https://www.erlang.org/) programming language and virtual machine.


b3nj5m1n:
Thank you!

PR #333, opened 2023-11-28
Add OpenJFX

Zweihander-Main:
[OpenJFX](https://github.com/openjdk/jfx) is the open source version of the JavaFX client application platform (equivalent to OpenJDK for the Java JDK). Programs which use it may create `~/.openjfx` as a cache folder. 

The ArchWiki also says it may create `~/.java/webview` so both the Java `userRoot` and JavaFX `cachedir` options are exported in `_JAVA_OPTIONS`. Users may already have the OpenJDK `userRoot` option set. As per the OpenJDK entry, disclaimer about applications ignoring the option has been added.

### DIFF
--- a/programs/openjfx.json
+++ b/programs/openjfx.json
@@ -1,0 +1,10 @@
+{
+    "files": [
+        {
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
Thank you!

PR #334, opened 2023-11-28
Add mtPaint

Zweihander-Main:
mtPaint is a graphics editor for creating icons and pixel art. It automatically creates a config file at `~/.mtpaint`. This shows a workaround using the system configuration file for mtPaint to manually set the user config file location. 

This still doesn't make it XDG compliant as it isn't setting it to an environment variable like `XDG_CONFIG_HOME` but at least gets the file out of the home directory.

### DIFF
--- a/programs/mtpaint.json
+++ b/programs/mtpaint.json
@@ -1,0 +1,11 @@
+{
+    "files": [
+        {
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
Thanks!

PR #340, opened 2023-12-05
mark zoom as not `movable`

robsdudeson:
moving `$HOME/.zoom` and setting `$SSB_HOME` doesn't really do anything since the next run of `zoom` re-creates the `$HOME/.zoom` dir with similar contents


b3nj5m1n:
Thank you!

PR #344, opened 2023-12-16
Add Kani

GrigorenkoPV:
https://github.com/model-checking/kani/blob/24ececc606d9a9978ea40f66a127c3b73ef0f921/src/setup.rs#L21-L25


b3nj5m1n:
Thanks!

PR #347, opened 2024-01-14
Added PCManFM, updated GIMP

tunalad:
Note that both PCManFM and GIMP used to store thumbnails in `$HOME/.thumbnails`. So now if the user has that directory, they will get messages for both apps

PCManFM: 
- https://github.com/lxde/libfm/issues/57

GIMP (cache and config):
- https://gitlab.gnome.org/GNOME/gimp/-/commit/483505f
- https://gitlab.gnome.org/GNOME/gimp/-/commit/60e0cfe

### DIFF
--- a/programs/gimp.json
+++ b/programs/gimp.json
@@ -1,9 +1,9 @@
 {
     "files": [
         {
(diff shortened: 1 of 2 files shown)


b3nj5m1n:
Thank you!

PR #352, opened 2024-01-27
Add a bunch of apps

### LINKED ISSUE #32: .android, .expo, .bubblewrap
There is no configuration for `.android`, which can be set using `ANDROID_USER_HOME`: https://developer.android.com/studio/command-line/variables

`.expo` is used by the [expo-cli](https://expo.dev/), which is An open-source platform for making universal native apps with React. Expo runs on Android, iOS, and the web.

`.bubblewrap` is used by [bubblewrap-cli](https://github.com/GoogleChromeLabs/bubblewrap), which helps developers to create a Project for an Android application that launches an existing Progressive Web App (PWAs) using a Trusted Web Activity.

### LINKED ISSUE #269: ~/.swiftmp
Generated by the Swift Package Manager and doesn't support the XDG specification. Related to https://github.com/apple/swift-package-manager/issues/6204.

### LINKED ISSUE #270: ~/.cocoapods
Generated by the CocoaPods and can be moved using the `CP_HOME_DIR` variable. Related to ["Can I change the default CocoaPods repositories folder?"](https://guides.cocoapods.org/using/faq.html#can-i-change-the-default-cocoapods-repositories-folder).

### LINKED ISSUE #272: ~/.tuist
Generated by Tuist and doesn't support the XDG specification. Related to https://github.com/tuist/tuist/discussions/5207.

### LINKED ISSUE #274: Moving .xinitrc is not warned to be potentially dangerous
The xinitrc should come with a warning that your window manager/desktop environment may not start if you move it. For me personally (BSPWM on Arch Linux), I use `startx`, which specifically looks for `$HOME/.xinitrc`. Note that in this situation, `startx` would have to be modified to look for `$HOME/.config/X11/xinitrc`, unless `$XDG_CONFIG_HOME` can be set before startx is called (This can be done by setting the XDG HOME variables in `.bash_profile` (or `.zprofile` for me). Most people probably use `.bashrc` for setting the variables though).

### LINKED ISSUE #283: ~/.ash_history
The shell ash provided by busybox uses `~/.ash_history` by default as history file. By exporting `HISTFILE=$XDG_STATE_HOME/ash/history` it can be moved to a xdg compilant location.

### LINKED ISSUE #289: Python recommendation returns an error when you exit ipython
I followed the recommendation for removing `.python_history`, and when I exit ipython, I get the following error message, and then ipython exits.

```console
Exception ignored in atexit callback: <function write_history at 0x7ff4c0737420>
Traceback (most recent call last):
  File "/etc/python/pythonrc", line 27, in write_history
    readline.write_history_file(history)
    ^^^^^^^^
NameError: name 'readline' is not defined

```

This seems like a weird result, given that the python file should be valid and readline should never be undefined, unless it's reading that file differently than normal python.

Python 3.11
IPython 8.14.0

### LINKED ISSUE #295: TLDR -  C client does have an environment variable to select cache directory location
``` json
 "help": "Currently tldr-pages do not have any specification for clients on where to cache tldr-pages. See the relevant [issue](https://github.com/tldr-pages/tldr/issues/876) and [specification](https://github.com/tldr-pages/tldr/blob/main/CLIENT-SPECIFICATION.md)\n\nOfficial tldr-nodejs-client, tldr-python-client or tldr-c-client do not support XDG Base Directories.\n\n**Use [Tealdeer](https://github.com/dbrgn/tealdeer), a rust client for tldr that supports XDG Base Directories instead.**\n”
```
[source](https://github.com/b3nj5m1n/xdg-ninja/blob/33b6491dd4561f134a89a22db5513df27219a15e/programs/tldr.json#L7C20-L7C20)

whilst the tldr c client doesnt use XDG it does look for the following environment variable to store the cache 

`TLDR_CACHE_DIR`

This can be used, for example,  to achieve 

``` bash
TLDR_CACHE_DIR=“$XDG_CACHE_HOME/tldrc”
```

``` c
gethome(void)
{
    char const *homedir = NULL;
    struct passwd *uid;

    if ((homedir = getenv("TLDR_CACHE_DIR")) != NULL)
        return homedir;

/* see code for the rest ....  */

    return homedir;
}
```
[source](https://github.com/tldr-pages/tldr-c-client/blob/ff0bd8ad12bbc2cc8f77449a04a3843dc7954691/src/utils.c#L29)

### LINKED ISSUE #297: [feature] some dot files to add for scriptkit app
### Script Kit

Script Kit runs on multiple platforms Linux, macOS, Windows and is used for automation scripting it’s a bit like macOS’s Hammerspoon but uses JavaScript based rather than Lua. Currently has no XDG support, however, does have some environment variables to change locations of files. I have suggested [XDG support ](https://github.com/johnlindquist/kit/issues/1320)

**Script Kit**
`~/.kit`
[http://www.scriptkit.com](http://www.scriptkit.com)
[https://github.com/johnlindquist/kit](https://github.com/johnlindquist/kit)
Environment Variable: `KIT`

**A "kenv" (Kit Environment)**
`~/.kenv `
[http://github.com/johnlindquist/kenv](https://github.com/johnlindquist/kenv)
seems to be the js scripts and modules for script kit 
environment variable for path: `KIT_DOTENV_PATH `

**Script Kits Nodejs**
`~/knode`
looks to be script kit’s node.js installation?
environment variable for path: `KNODE `

> What are Kit.app, kit, and kenvs?
> 
> Kit.app - The Kit.app provides a UI for your scripts. The app is "script-driven" meaning that every time you launch the app, you're really launching a script. The main menu, even though complex, is a script you could write.
> 
> kit - "kit" is the sdk of Script Kit
> 
> A bundle of JavaScript common libs wrapped by an API to make writing scripts easier (get, download, replace, outputFile, etc)
> APIs for interacting with your OS (edit, focusTab, say, notify, etc)
> APIs for interacting with Kit.app and Terminal (arg, env, etc)
> Scripts and utils for app setup, managing kenvs, parsing scripts, etc
> kenvs - Kit Enviroments (AKA "kenv") are directories that contain a "scripts" directory. If you point "kit" at a "kenv", kit will parse the scripts and give you tools to simplify running and managing them.

### LINKED ISSUE #300: xdg-utils update
https://github.com/b3nj5m1n/xdg-ninja/blob/30fd34683ed4cbc3f59eb7c7784fff1e2a7eb501/programs/xdg-utils.json#L2-L9

[That merge request just got merged](https://gitlab.freedesktop.org/xdg/xdg-utils/-/merge_requests/22)

### LINKED ISSUE #315: ~/.emacs.d
From: https://www.gnu.org/software/emacs/manual/html_node/emacs/Find-Init.html
>  Emacs can also look in an XDG-compatible location for init.el, the default is the directory ~/.config/emacs. This can be overridden by setting XDG_CONFIG_HOME in your environment, its value replaces ~/.config in the name of the default XDG init file. However ~/.emacs.d, ~/.emacs, and ~/.emacs.el are always preferred if they exist, which means that you must delete or rename them in order to use the XDG location.

So you can safely do:
```bash
mv ~/.emacs.d $XDG_CONFIG_HOME/emacs
```

### LINKED ISSUE #341: zsh history and compdump locations must be created manually
Per the instructions in the zsh file, relocating the compdump and history files to the xdg dirs did not work for me until I manually created the `zsh` dir within the xdg dirs: 
https://github.com/b3nj5m1n/xdg-ninja/blob/74dd300df17d2af5e158070edd79d31491f51ef0/programs/zsh.json#L8-L27

Is this a general issue that should be addressed within the file's instructions or is it just an issue with my machine?

### LINKED ISSUE #342: $ANDROID_HOME refers to SDK installation which is ~/Android and not ~/.android which is a config folder.
Output of xdg-ninja ->
```
[android-studio]: $HOME/.android

  Export the following environment    
  variables:                          
                                    
    export                            
  ANDROID_HOME="$XDG_DATA_HOME"/android
```

---------

Source - https://developer.android.com/tools/variables#envar

<table>
<tr>
 <td>ANDROID_HOME
 <td>Sets the path to the SDK installation directory. Once set, the value does not typically change and can be shared by multiple users on the same machine. ANDROID_SDK_ROOT, which also points to the SDK installation directory, is deprecated. If you continue to use it, Android Studio and the Android Gradle plugin will check that the old and new variables are consistent. 
<tr>
 <td>ANDROID_USER_HOME
 <td>Sets the path to the user preferences directory for tools that are part of the Android SDK. Defaults to $HOME/.android/.

Some older tools, such as Android Studio 4.3 and earlier, do not read ANDROID_USER_HOME. To override the user preferences location for those older tools, set ANDROID_SDK_HOME to the parent directory you would like the .android directory to be created under. 
</table>

So ANDROID_USER_HOME corresponds to $HOME/.android/ & ANDROID_HOME corresponds to $HOME/Android/

### LINKED ISSUE #349: dotnet now respects DOTNET_CLI_HOME environment variable to define the location of settings
https://github.com/b3nj5m1n/xdg-ninja/blob/c9f7b9d48cfdae8d7ec6f913b549da4b634efd75/programs/dotnet.json#L3-L7

According to a [comment in the linked issue](https://github.com/dotnet/sdk/issues/8678#issuecomment-1509276940) dotnet now respects `$DOTNET_CLI_HOME` environment variable.

I'm unsure what the "right" way to configure it is though since the .dotnet directory contains all kinds of garbage (much of it belongs in `$XDG_CACHE_HOME` but I think there can also be some actual configuration files and possibly state).

There's also a newer issue tracking XDG support: https://github.com/dotnet/sdk/issues/10390

flexagoon:
closes #32
closes #269 
closes #270 
closes #272
closes #274 
closes #283 
closes #289 
closes #295 
closes #297 
closes #300 
closes #315 
closes #341 
closes #342
closes #349


sertonix:
The directory for ash needs to be created manually too.

flexagoon:
@b3nj5m1n could you please review this?

b3nj5m1n:
This is awesome, thank you so much!

sertonix:
@b3nj5m1n A lot of things have changed since the last release. could you tag a new release?

b3nj5m1n:
> @b3nj5m1n A lot of things have changed since the last release. could you tag a new release?

Here you go: https://github.com/b3nj5m1n/xdg-ninja/releases/tag/v0.2.0.2

As I've said before, the releases were always meant for the Haskell portion of the codebase, but I've been wanting to switch to a new versioning system with automated releases for the program configs for months now and haven't had the time, so this is definitely better for now.

flexagoon:
@b3nj5m1n it's pretty easy to set up automatic releases with GitHub actions, I can help with that if you want

b3nj5m1n:
> @b3nj5m1n it's pretty easy to set up automatic releases with GitHub actions, I can help with that if you want

@flexagoon That would be awesome, I think we really just need to generate the changelogs, increase the semver by one (I think we should start using a new versoning scheme for the program configs) and decide when to generate new releases (either something like once a month or every 5 commits or something like that).

It's definitely not hard, I just have very little time with uni, so I never get around to doing it.

PR #355, opened 2024-02-01
Add solution for renpy

vigress8:
Ren'Py supports the `RENPY_PATH_TO_SAVES` variable to set its save directory.


b3nj5m1n:
Thank you!

PR #357, opened 2024-02-07
Fix link for bun

thunfisch987:
The [linked issue](https://github.com/oven-sh/bun/issues/696) was closed as a duplicate of https://github.com/oven-sh/bun/issues/1678 so i updated the link

i also moved the dot after the link because in (at least my terminal, using konsole and zsh) clicking on the link would take one to .../issues/696. instead of .../issues/696 which didn't work

### DIFF
--- a/programs/bun.json
+++ b/programs/bun.json
@@ -3,7 +3,7 @@
         {
             "path": "$HOME/.bun",
             "movable": false,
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
Thank you!

Issue #368, opened 2024-02-21
Detection missing for jack


n-hebert:
There's a ~/.jackdrc

PR #369, opened 2024-02-21
Create jack2.json to detect ~/.jackdrc

n-hebert:
Hard-coded as shown at https://github.com/jackaudio/jack2/blob/develop/common/JackServerGlobals.cpp#L166


n-hebert:
Closes #368

PR #382, opened 2024-03-10
Use $XDG_* in zsh help, rather than $HOME/.config

LucasFA:
Move the file to $XDG_* instead of $HOME/.config

Unless I am missing something, no need to specify $HOME


b3nj5m1n:
Thanks!

PR #384, opened 2024-03-12
Add new-line character

leoheitmannruiz:
Thanks for xdg-ninja!

This PR adds a new-line character, so that `xdg-ninja -h` executes as I would expect.
Please take a look at the current output. I'm not sure how to describe what's going on.
Well, there's a character that's not supposed to be there..

See `printf "%d" 0x20` versus `printf "%d\n" 0x20`.

Perhaps I'm missing something though :)


b3nj5m1n:
I've somehow never noticed that, thank you!

PR #391, opened 2024-04-10
Add Sweet Home 3D

Zweihander-Main:
[Sweet Home 3D](https://www.sweethome3d.com/) is a 3D interior design application. Programs which use it may create `~/.eteks/sweethome3d` as a data folder.


b3nj5m1n:
Thank you!

PR #394, opened 2024-04-21
Add zsh_sessions entry to zsh

MeanderingProgrammer:
Thanks for centralizing all of this information into code, this tool is awesome!

## Details

Some info on the .zsh_sessions file can be found here: https://apple.stackexchange.com/questions/427561/macos-zsh-sessions-zsh-history-and-setopt-append-history

It only gets generated on MacOS when using the default terminal and can be disabled in the user zshenv file.

### DIFF
--- a/programs/zsh.json
+++ b/programs/zsh.json
@@ -1,6 +1,11 @@
 {
     "files": [
         {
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
Thanks!

PR #397, opened 2024-04-25
Add flake-compat

### LINKED ISSUE #396: Add flake-compat
Could you please add [flake-compat](https://github.com/edolstra/flake-compat) to the repository, so it could be used from the systems without flakes enabled

WeetHet:
Fixes #396


WeetHet:
It seems this is not something maintainers need or want, so I'm closing this for now

PR #400, opened 2024-05-05
Update and reformat `README`

KaBankz:
I've updated the install section for Homebrew to install from the git HEAD, instead of the latest release tag. ref: #204

I've also reformatted and changed the verbage of the README to be more consice and clear.
The README now follows `markdownlint` rules, and best practices for markdown.

I made sure to not change the actual meaning of the README's contents, just the way it was presented.

If you feel like the changes are not appropriate, please let me know and I'll revert them, keeping only the updated Homebrew install instructions.

If you have any questions or suggestions, please let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,9 +28,24 @@ Clone the repository somewhere, then run the _./xdg-ninja.sh_ script.
 
 This will run every test in the default configuration.
 
-### Installing with Homebrew
+### [Homebrew](https://brew.sh)
 
-To install xdg-ninja with [Homebrew](https://brew.sh), run `brew install xdg-ninja` to install the script and all of its dependencies, then run the `xdg-ninja` command.
+> [!NOTE]
+> Due to how `xdg-ninja` is developed, releases are not cut, so `brew` ships a stale version, therefore you have to install and upgrade `xdg-ninja` from the git HEAD. ref: [#204](https://github.com/b3nj5m1n/xdg-ninja/issues/204)
+>
+> Homebrew will not upgrade `xdg-ninja` when running a generic `brew upgrade`, you must specifically upgrade `xdg-ninja` from the git HEAD, _see below_
+
+Install:
+
+```sh
+brew install xdg-ninja --HEAD
+```
+
+Upgrade:
+
+```sh
+brew upgrade xdg-ninja --fetch-HEAD
+```
 
 ## Dependencies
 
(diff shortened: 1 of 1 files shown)


KaBankz:
Thanks for the feedback, I'm glad you like the changes!

> I'm a little hesitant to put the big homebrew section before the instructions for manual installation, since most people are probably not using homebrew.

No problem, I'll bump the manual installation instructions to the top.

Any thoughts on adding the commands to that as well? So, something like this:

````md
### Manual Installation

Clone the repository, then run the [`./xdg-ninja.sh`](./xdg-ninja.sh) script.

```sh
git clone https://github.com/b3nj5m1n/xdg-ninja
cd xdg-ninja
./xdg-ninja.sh
```

This will run every test in the default configuration.
````

Also, what about including a section like this, under the rest of the installation methods:

```md
### Other Package Managers

`xdg-ninja` is avaliable in many package managers.

The full list is available on the [repology page](https://repology.org/project/xdg-ninja/versions).

Follow the instructions for your package manager to install `xdg-ninja`.
```

> I also think the "Available commands" section for xdgnj should be higher up, I think putting it right after the "You can download the xdgnj executable from the releases page." line makes more sense, since (I'm assuming) most people are using the executable from the releases page, and not building it themselves.

I purposefully moved it lower because it flows better:

Installation Methods -> Available Commands

Rather than:

Installation Method 1 -> Available Commands -> Installation Method 2 -> etc...

But, if that's your preference, then I can move it back up.

Thoughts on adding instructions on how to download from the releases? Something like this:

````md
#### Prebuilt Binaries

> [!IMPORTANT]
> The binaries only run on x86_64 Linux systems.

```sh
curl -fsSL -o xdgnj https://github.com/b3nj5m1n/xdg-ninja/releases/latest/download/xdgnj
chmod +x xdgnj
```
````

Thanks!

b3nj5m1n:
I'm on board with all of that!

KaBankz:
Awesome, just to clarify, you want the "Available commands" section moved back up or keep it as it is?

b3nj5m1n:
I'd prefer it higher up, I totally get where you're coming from, I just think most users would probably like to get an overview of what the utility is actually good for before going through specifics.

KaBankz:
Absolutely, check it now, is it high enough, or should I move it higher?

Also, if you have anything else you want added removed or changed, let me know.

b3nj5m1n:
Awesome, thank you very much again!

PR #402, opened 2024-05-07
Add lyrics entry to ncmpcpp

Leenuus:
# Commit detail
By default, ncmpcpp creates `~/.lyrics` to store lyrics, other mpd clients use this directory too. For ncmpcpp users, this behavior can be changed by modifying ncmpcpp config file. But it may not work if other mpd clients exists.
see `man ncmpcpp`, search `lyrics` to get more information.

# Conflicts
However, this directory seems used by an existing program named `lyrics` in xdg-ninja. So currently, when user run xdg-ninja, both entries appear, which may cause some confusion.

lyrics


b3nj5m1n:
Thanks!

PR #410, opened 2024-05-31
Update parallel.json

fedemp:
Support for XDG mentioned in https://savannah.gnu.org/bugs/?51294

Tested with parallel version 20230822


fedemp:
My bad. `parallel` does indeed create the folder again after deleting it. Moving it to .config only removes the citation notice.

PR #412, opened 2024-06-05
Add neverball

JoelleJS:
Project: https://github.com/Neverball/neverball

Unfortunately the configuration isn't movable, see https://github.com/Neverball/neverball/pull/184


b3nj5m1n:
Thanks!

PR #413, opened 2024-06-05
Add ICAClient aka Citrix Workspace App

JoelleJS:
No mention of a configurable directory in the documentation so presumed to be hardcoded

https://docs.citrix.com/en-us/citrix-workspace-app-for-linux/app-preferences


b3nj5m1n:
Thank you!

PR #414, opened 2024-06-05
Add Red Hat telemetry data

JoelleJS:
Add info on the `.redhat` directory containing Red Hat telemetry data.

I reported the issue with Red Hat, but for now it seems that making the directory movable is not a priority. See https://github.com/redhat-developer/vscode-redhat-telemetry/issues/51.


b3nj5m1n:
Thanks!

PR #415, opened 2024-06-16
Fix `.tmux` output suggesting incorrect target path.

orendain:
The previous `.tmux` output instructed users to update their TPM run command to `run ~/.local/share/...`, which is typically represented by `XDG_DATA_HOME`. However, the directions also told users to then move the corresponding `.tmux` directory to `XDG_CONFIG_HOME/tmux`.

The two halves of the instructions were conflicting.

To resolve this, this commit fixes the second half of the instructions with the correct target path (i.e., `XDG_DATA_HOME`).

### DIFF
--- a/programs/tmux.json
+++ b/programs/tmux.json
@@ -9,7 +9,7 @@
         {
             "path": "$HOME/.tmux",
             "movable": true,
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
Thanks!

PR #416, opened 2024-06-17
remove .gitignore rule

intiluha:
$HOME/.config/git/ignore file is for listing global ignore rules for all user repositories. 
$HOME/.gitignore file is for listing ignore rules for $HOME repository (if you have any). 
They are not the same


b3nj5m1n:
Thank you!

PR #417, opened 2024-06-27
Use native readline function for new history file

ItsDrike:
In commit 07bcbe565775da519060cef3f381ad9d10d70665 an issue described in https://github.com/python/cpython/issues/105694 was solved by changing the pythonrc script and adding:

```python
with open(history,\"w\") as f:
    f.write(\"_HiStOrY_V2_\" + \"\\n\\n\") # breaks on macos + python3 without this.
```

Even though this does indeed solve the issue, it's really ugly code-wise, and there is a built-in way to do this in the `readline` library, with:

```python
readline.write_history_file(str(history)) # breaks on macos + python3 without this.
```

This PR just edits the script to use that instead.


b3nj5m1n:
Thanks!

PR #419, opened 2024-06-30
add nyx.json

heather7283:
First, create a _$XDG_CONFIG_HOME/nyxrc_ file with the following content:

```
data_directory ~/.cache/nyx
```
Note that you can't use variable substitution inside nyxrc, so you will have to hardcode your _XDG_CACHE_HOME_.
Alternatively, disable caching to disk:

```
data_directory disabled
```
Finally, in your shell startup file:

```sh
alias nyx="nyx --config $XDG_CONFIG_HOME/nyxrc"
```
Source: https://nyx.torproject.org/#configuration

### DIFF
--- a/programs/nyx.json
+++ b/programs/nyx.json
@@ -1,0 +1,10 @@
+{
+    "name": "nyx",
+    "files": [
+        {
+            "path": "$HOME/.nyx",
+            "movable": true,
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
Thank you!

PR #421, opened 2024-07-15
Distinguish movable vs unmovable ~/.aws paths

sparr:
After following the existing instructions for ~/.aws I still got the same help because the un-movable ~/.aws/cli and sso still existed. This PR gives more specific instructions/info for the different paths.


b3nj5m1n:
Thank you!

PR #422, opened 2024-07-17
Fix fontconfig base dirs

### LINKED ISSUE #388: [BUG] xdg-ninja says to move ~/.fonts.conf to $XDG_DATA_HOME instead of $XDG_CONFIG_HOME
https://github.com/b3nj5m1n/xdg-ninja/pull/218,  says that it should be $XDG_CONFIG_HOME.

Also please look at this ->

xdg-ninja output regarding ~/.fonts directory
```
<?xml version="1.0"?><!DOCTYPE fontconfig SYSTEM "fonts.dtd">
<fontconfig>
 <dir>~/.fonts</dir>
</fontconfig>
```

~/.fonts.conf
```
<?xml version="1.0"?><!DOCTYPE fontconfig SYSTEM "fonts.dtd">
<fontconfig>
 <dir>~/.fonts</dir>
</fontconfig>
```

so after moving both, .conf needs to be edited? If yes, then a note indicating that would be nice.

baschi29:
According to https://www.freedesktop.org/software/fontconfig/fontconfig-user.html :

1) Fixes an typo in #218: It did not add `$XDG_CONFIG_HOME` but `$XDG_DATA_HOME` as the destination for `${HOME}/.fonts.conf`.
Also the possibility of one needing to change the `<dir>` element in the config is now mentioned.

2) While investigating I also noticed that `${HOME}/.fontconfig` should be moved to `$XDG_CACHE_HOME/fontconfig` instead of `${XDG_DATA_HOME}/fontconfig`

This PR closes #388

### DIFF
--- a/programs/fontconfig.json
+++ b/programs/fontconfig.json
@@ -4,12 +4,12 @@
         {
             "path": "$HOME/.fontconfig",
             "movable": true,
-            "help": "Supported\n\nThe file $HOME/.fontconfig can be moved to $XDG_DATA_HOME/fontconfig.\n"
+            "help": "Supported\n\nThe file $HOME/.fontconfig can be moved to $XDG_CACHE_HOME/fontconfig.\n"
         },
         {
             "path": "$HOME/.fonts.conf",
             "movable": true,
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
Thanks!

PR #423, opened 2024-07-17
Add nrfutil

h7x4:
Project: [nRF Util](https://www.nordicsemi.com/Products/Development-tools/nRF-Util)

Not entirely obvious from the docs that this envvar exists, might be a good addition.


b3nj5m1n:
Thank you!

PR #429, opened 2024-08-14
Add rvm

joshuanianji:
Sources:

- Github: https://github.com/rvm/rvm/blob/master/examples/rvmrc#L10-L13
- Official site (they are a little unclear about the role of rvm_path): https://rvm.io/rvm/configuration

Note that the environment variable is lowercase


b3nj5m1n:
Thanks!

PR #430, opened 2024-08-14
Add zsh-evalcache

joshuanianji:
Reference: https://github.com/mroth/evalcache/tree/master?tab=readme-ov-file#options


b3nj5m1n:
Thank you!

PR #433, opened 2024-09-25
Make PYTHONSTARTUP script respect PYTHON_HISTORY on Python <3.13

LunarLambda:
This allows the user to set `PYTHON_HISTORY` as the single source of truth on python versions prior to 3.13. As an extra bit of redundancy, the script now checks the python version too, making it a no-op on Python 3.13 and later.


b3nj5m1n:
Thank you!

PR #443, opened 2024-11-08
Missing three $ in conda help

clyhtsuriva:
Hi, this is a minor PR to fix a small issue in the conda help text. I noticed that there were three missing dollar signs ($).


b3nj5m1n:
Thanks!

PR #448, opened 2024-12-31
Update .Renviron to use curly brackets for Environment Variables

marcingold22:
I utilized this tool to clean my home environment, and everything worked perfectly, with one exception: upon starting RStudio, a `$XDG_DATA_HOME` folder was created. Yes, the full name not the value of the variable.

Upon investigation, I discovered that the `.Renviron` file requires environment variables to be encapsulated in curly brackets.

This pull request addresses this issue by updating the `.Renviron` file accordingly.

### DIFF
--- a/programs/R.json
+++ b/programs/R.json
@@ -4,7 +4,7 @@
         {
             "path": "$HOME/R",
             "movable": true,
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
Thank you!

PR #457, opened 2025-02-17
Add missing underscore to jupyter entry

claydugo:
Caused me a bit of of confusion when learning the output format of this tool.


b3nj5m1n:
Thank you!

PR #460, opened 2025-03-06
Add GHCi

ehamberg:
Seems to have been supported for quite some time: https://downloads.haskell.org/ghc/9.4.1/docs/users_guide/ghci.html#the-ghci-files


b3nj5m1n:
Thank you!

PR #462, opened 2025-03-10
Move `--absolute` before `--dir`

ItsProfessional:
Move `--absolute` before `--dir`. This is required otherwise the `--absolute` flag doesn't have an effect.


b3nj5m1n:
Thank you!

PR #463, opened 2025-03-20
Update TigerVNC documentation

fox8091:
TigerVNC uses XDG Base Directory Specification paths as of PRs [#1737](https://github.com/TigerVNC/tigervnc/pull/1737) and [#1760](https://github.com/TigerVNC/tigervnc/pull/1760), update documentation to match.

### DIFF
--- a/programs/tigervnc.json
+++ b/programs/tigervnc.json
@@ -3,7 +3,7 @@
         {
             "path": "$HOME/.vnc",
             "movable": true,
-            "help": "Supported by default.\n\nNewer versions of tigervnc automatically create directories at correct locations, consult `man vncviewer` to see where to move files.\n**WARNING: Failure to move files to the correct location can result in locking out VNC access**"
+            "help": "Supported by default.\n\nAll versions above 1.14.0 will automatically create directories at correct locations, consult `man vncviewer` to see where to move files.\n**WARNING: Failure to move files to the correct location can result in locking out VNC access**"
         }
     ],
     "name": "tigervnc"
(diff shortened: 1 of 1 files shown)


programs/tigervnc.json
```diff
@@ -2,8 +2,8 @@
     "files": [
         {
             "path": "$HOME/.vnc",
-            "movable": false,
-            "help": "Currently unsupported.\n\n_Relevant issue:_ https://github.com/TigerVNC/tigervnc/issues/1195\n"
+            "movable": true,
+            "help": "Supported by default.\n\nNewer versions of tigervnc automatically create directories at correct locations, consult `man vncviewer` to see where to move files.\n**WARNING: Failure to move files to the correct location can result in locking out VNC access**"
```
kamack38, line 6:
```suggestion
            "help": "Supported by default.\n\nAll versions above 1.14.0 will automatically create directories at correct locations, consult `man vncviewer` to see where to move files.\n**WARNING: Failure to move files to the correct location can result in locking out VNC access**"
```

kamack38 (changes requested):
I think it should be specified from what version the XDG Base Dir spec is supported.

programs/tigervnc.json
kamack38: I don't know whether it should not be changed to 1.13.1, since the change was introduced in [1.14.0](https://github.com/TigerVNC/tigervnc/releases/tag/v1.14.0).

b3nj5m1n:
Thanks!

PR #466, opened 2025-04-14
Xauthority XDG Desktop Portal warning

tunalad:
I might have worded the message a bit poorly, but in short, we're supposed to force an environment variable to the systems service. In this case, the `XAUTHORITY` environment variable.

If anyone can word this any better and shorter, please do xd

edit: sorry for all the extra unrelated commits below


b3nj5m1n:
Thank you!

PR #492, opened 2025-08-01
feat: add tip for bash history

Kobu:
I was wondering why my history file was not being used when using reverse search, because I added `export HISTFILE="${XDG_STATE_HOME}"/bash/history` as was instructed. Turns out the file needs to exist and is not created automatically. 

I added a simple helper message that points this out:
image


b3nj5m1n:
Thanks!

PR #493, opened 2025-08-04
Add support for 1Password CLI.

jdwhite:
While not XDG compliant, the `.op` directory it creates by default can be relocated and its location specified at runtime.


b3nj5m1n:
Thank you!

PR #499, opened 2025-08-15
link to the "why"

meonkeys:
It took me a while to figure out why xdg-ninja exists. I hope this is useful & helpful for others.

Credit to @chuckwolber for the helpful link.

Best and thanks,
-Adam


b3nj5m1n:
Good idea, thank you!

PR #500, opened 2025-08-20
Update aspell/hunspell entries; add other new programs

Nightfirecat:
The first commit helps match dictionary/repl files created by aspell and hunspell. The commits following that provide matchers for the following programs:
* [abcde](https://abcde.einval.com/wiki/)
* [reportbug](https://wiki.debian.org/reportbug)
* [RuneLite](https://runelite.net)
* [shellcheck](https://www.shellcheck.net/)
* [xmodmap](https://gitlab.freedesktop.org/xorg/app/xmodmap)

### DIFF
--- a/programs/abcde.json
+++ b/programs/abcde.json
@@ -1,0 +1,10 @@
+{
+    "files": [
+        {
+            "path": "$HOME/.abcde.conf",
+            "movable": false,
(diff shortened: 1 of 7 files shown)


b3nj5m1n:
Thank you!

PR #505, opened 2025-09-10
Added support for NetExec

sdushantha:
NetExec add support for XDG Base Directories in https://github.com/Pennyw0rth/NetExec/pull/649


b3nj5m1n:
Thank you!

PR #513, opened 2025-11-17
Added support for jq module directory

steffennilsen:
Added support for the hardcoded jq module directory `$HOME/.jq`. This is unfortunately used for both the default module directory AND a global macro file. The workaround alias suggested only works for the module directory use case and not for the global script use case, thus marking it as `movable: false`


b3nj5m1n:
Thanks!

PR #514, opened 2025-11-21
firefox now supports ~/.config

killertofus:


programs/firefox.json
```diff
@@ -2,9 +2,8 @@
     "files": [
         {
             "path": "$HOME/.mozilla",
-            "movable": false,
-            "help": "Currently unsupported.\n\n_Relevant issue:_ https://bugzilla.mozilla.org/show_bug.cgi?id=259356\n"
-        }
+            "movable": True,
```
gcp: With some caveats, see https://bugzilla.mozilla.org/show_bug.cgi?id=259356#c230

gcp (changes requested):

programs/firefox.json
```diff
@@ -2,9 +2,8 @@
...
+             "help": "Supported\n\nThe file $HOME/.mozilla can be moved to $XDG_CACHE_HOME/mozilla.\n"}
```
gcp: Didn't you mean to write `$XDG_CONFIG_HOME/mozilla`?

gcp (changes requested):

programs/firefox.json
```diff
@@ -2,9 +2,8 @@
     "files": [
         {
             "path": "$HOME/.mozilla",
-            "movable": false,
-            "help": "Currently unsupported.\n\n_Relevant issue:_ https://bugzilla.mozilla.org/show_bug.cgi?id=259356\n"
-        }
+            "movable": False,
+             "help": "Supported\n\nThe file $HOME/.mozilla can be moved to $XDG_CONFIG_HOME/mozilla.\n"}
```
gcp: I mean it sounds weird to say `movable: False` and then to say you can just move it. It's movable but with risk of some minor breakage.

gcp (approved):

programs/firefox.json
```diff
@@ -2,9 +2,8 @@
     "files": [
         {
             "path": "$HOME/.mozilla",
-            "movable": false,
-            "help": "Currently unsupported.\n\n_Relevant issue:_ https://bugzilla.mozilla.org/show_bug.cgi?id=259356\n"
-        }
+            "movable": True,
+             "help": "Supported\n\nThe file $HOME/.mozilla can be moved to $XDG_CONFIG_HOME/mozilla.\n with risk of minor breakage"}
```
jneidel:
```suggestion
             "help": "Supported from version 147.\n\nCannot be migrated consistently. Try migration, if that fails backup and remove $HOME/.mozilla and start firefox to have a new profile be created in $XDG_CONFIG_HOME/mozilla.\n\n_Relevant issue:_ https://bugzilla.mozilla.org/show_bug.cgi?id=259356\n"}
```

Just moving it did not work for me.
From the [bug thread](https://bugzilla.mozilla.org/show_bug.cgi?id=259356#c228):

> If there is $HOME/.mozilla/firefox it will continue with legacy behavior. The preferred way to move to the new behavior would be to ensure Sync is setup, get rid of the legacy directory (make a backup just in case) and starting again the browser. A new profile should be created in the XDG_CONFIG_HOME hierarchy and then you can bring back things. Unfortunately, handling more automatic migration is more error prone and with experience it was hardly reliable.

jneidel (approved):

programs/firefox.json
jneidel: I would set it to false to reflect the complication that might arise.

```suggestion
            "movable": false,
```

If you want true, correct it to

```suggestion
            "movable": true,
```

jneidel (changes requested):

b3nj5m1n:
Thanks very much!

PR #525, opened 2025-12-21
feat(programs/codex): init

fdnt7:
The [official configuration documentation](https://github.com/openai/codex/blob/main/docs/config.md) states that the `CODEX_HOME` environment value where it defaults to `~/.codex` will be where the configuration file `$CODEX_HOME/config.toml`, logs and other Codex-related information are stored.

### DIFF
--- a/programs/codex.json
+++ b/programs/codex.json
@@ -1,0 +1,10 @@
+{
+    "files": [
+        {
+            "path": "$HOME/.codex",
+            "movable": true,
+            "help": "Export the following environment variables:\n\n```bash\nexport CODEX_HOME=\"$XDG_CONFIG_HOME\"/codex\n```\n"
+        }
+    ],
+    "name": "codex"
+}
--- a/programs/codex.json
+++ b/programs/codex.json
@@ -3,7 +3,7 @@
         {
             "path": "$HOME/.codex",
             "movable": true,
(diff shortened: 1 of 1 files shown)


mise42:
They also place data and state in `$CODEX_HOME`, so it's not appropriate to make `$CODEX_HOME=$XDG_CONFIG_HOME/codex`

programs/codex.json
```diff
@@ -0,0 +1,10 @@
+{
+    "files": [
+        {
+            "path": "$HOME/.codex",
+            "movable": true,
+            "help": "Export the following environment variables:\n\n```bash\nexport CODEX_HOME=\"$XDG_CONFIG_HOME\"/codex\n```\n"
```
b3nj5m1n:
```suggestion
            "help": "Export the following environment variables:\n\n```bash\nexport CODEX_HOME=\"$XDG_CONFIG_HOME\"/codex\n```\nWarning: also contains data that isn't configuration.\n"
```

b3nj5m1n:
Thanks!

rea1shane:
It seems this configuration is no longer correct. I’ve set `CODEX_HOME`, but CodeX is still writing files to `~/.codex`. Now there are files in both `~/.codex` and `~/.config/codex`, and I can’t tell which one it’s actually using.

UPDATE: Looking at the modification times of the files under these two paths, it seems that they are both being used.

PR #526, opened 2025-12-25
Add support for ly log file

ABDsheikho:
ly is a Display manager (login manager) that stores its log file in the _home_ directory. the ly log file can be easily moved by editing its config file in the `/etc/ly/config.ini` file.

Since the ly config file is located in `/etc` directory, it doesn't support shell variables like `$HOME` or `$XDG_STATE_HOME`, but it can support it natively by hard coding `.local/state/` in its config instead of `$XDG_STATE_HOME`.

This PR adds a `programs/ly.json` file, which contains instructions on how to move it.

### DIFF
--- a/programs/ly.json
+++ b/programs/ly.json
@@ -1,0 +1,10 @@
+{
+    "name": "ly",
+    "files": [
+        {
+            "path": "$HOME/ly-session.log",
+            "movable": true,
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
Hi, while going through the backlog of PRs I first merged #547, sorry about that and thank you for your efforts!

ABDsheikho:
No problem, but my PR was 4 months ago, and in that time `ly` moved their log file to `.local/state` dir, which made both PR unnecessary.

Issue #506, opened 2025-10-04
`~/.renpy/persistent`


jrom99:
Renpy has another environment variable that specifies persistent files (similar to save files) `RENPY_MULTIPERSISTENT`.

PR #528, opened 2026-01-02
Fix renpy.json

hizani:
help says it is movable


jrom99:
There is also another variable, see https://github.com/b3nj5m1n/xdg-ninja/issues/506

b3nj5m1n:
Thank you!

PR #479, opened 2025-05-31
Add files for git & gitkraken + update php

LuckyWindsck:
Hi, thank you for this awesome project.

I've updated a few files in `programs/`, the following are some references to the change.

- git
    - .gitconfig: https://git-scm.com/docs/git-config#SCOPES
    - .git-credentials: https://git-scm.com/docs/git-credential-store#FILES
    - .gitignore: https://git-scm.com/docs/gitignore#_synopsis
    - .gitk: https://git-scm.com/docs/gitk#_files
- gitkraken
    - .gk
        - Issue: https://github.com/gitkraken/vscode-gitlens/issues/3660
        - PR: https://github.com/gitkraken/vscode-gitlens/pull/3904
        - Implementation (v17.1.1): https://github.com/gitkraken/vscode-gitlens/blob/v17.1.1/src/env/node/gk/localSharedGkStorageLocationProvider.ts#L33
- php
    -  .php_history: https://www.php.net/manual/en/migration84.new-features.php#migration84.new-features.readline

PR #529, opened 2026-01-07
Fix .gitignore rule description

sliva0:
`.gitignore` rule initially was not there, then it was added in #215 with incorrect description, removed in #416 with a good reasoning, and then added again in #479 with the same text as in #215.

To avoid third reintroduction of this rule I suggest marking it as not movable instead and adding more descriptive description to explain what's the difference between `~/.gitignore` and `~/.config/git/ignore`.

### DIFF
--- a/programs/git.json
+++ b/programs/git.json
@@ -18,8 +18,8 @@
         },
         {
             "path": "$HOME/.gitignore",
-            "movable": true,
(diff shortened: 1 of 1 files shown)


b3nj5m1n:
Yeah sorry about that, thanks very much!

PR #532, opened 2026-01-26
Add ripgrep

### LINKED ISSUE #531: Support ripgrep
[ripgrep](https://github.com/BurntSushi/ripgrep) is a grep-like file searcher. It can support a configuration file through setting an environment variable, so there's potential for someone to have manually set it already. I still think it's worth checking anyway.

NoahTheDuke:
Closes #531

### DIFF
--- a/programs/ripgrep.json
+++ b/programs/ripgrep.json
@@ -1,0 +1,10 @@
+{
+    "files": [
+        {
+            "path": "$HOME/.ripgreprc",
+            "movable": true,
+            "help": "Export the following environment variable:\n\n```bash\nexport RIPGREP_CONFIG_PATH=\"$XDG_CONFIG_HOME/.ripgreprc\"\n```\nThen move the config file to _$XDG_CONFIG_HOME/.ripgreprc_.\n"
+        }
+    ],
+    "name": "ripgrep"
+}
--- a/programs/ripgrep.json
+++ b/programs/ripgrep.json
@@ -3,7 +3,7 @@
         {
             "path": "$HOME/.ripgreprc",
             "movable": true,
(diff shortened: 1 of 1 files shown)


programs/ripgrep.json
```diff
@@ -0,0 +1,10 @@
+{
+    "files": [
+        {
+            "path": "$HOME/.ripgreprc",
+            "movable": true,
+            "help": "Export the following environment variable:\n\n```bash\nexport RIPGREP_CONFIG_PATH=\"$XDG_CONFIG_HOME/.ripgreprc\"\n```\nThen move the config file to _$XDG_CONFIG_HOME/.ripgreprc_.\n"
```
lgarron: Saw this pass by, and I think `$XDG_CONFIG_HOME/.ripgreprc` is not an ideal recommendation:

- Using a leading `.` inside an XDG folder is pretty uncommon — I can't find a single example of that on my system, out of over 80 paths. (Making it a dot dir makes it liable to overlook by accident, if someone happens to be listing paths in a way that does not include nested hidden folders — even `ls ~/.config`.)
- It's not unheard-of to place a file directly under an XDG path root, but the vast majority of programs I've encountered use at least a single folder as a namespace.

Since `rg` doesn't have a default convention for this, I think a recommendation matching other programs would be good, e.g. something like:

```
$XDG_CONFIG_HOME/ripgrep/config
```

In particular, this closely resembles `git`.

(Arguably something like `$XDG_CONFIG_HOME/ripgrep/default-flags` might be clearer, but `rg` does call its own var a config path.)

NoahTheDuke: sure, that works for me!

b3nj5m1n:
Thank you!

PR #533, opened 2026-01-26
Add rule for Clojure ~/.gitlibs directory

greglook:
The Clojure programming language allows referencing dependencies via a Git repository URL, and those dependencies are cached using [tools.gitlibs](https://github.com/clojure/tools.gitlibs). By default, that puts the libraries in `~/.gitlibs`, but this can be overridden with an environment variable.


b3nj5m1n:
Thanks!

PR #535, opened 2026-01-31
Add triton

engdyn:
`export TRITON_HOME="$XDG_CACHE_HOME"` will cause triton to create the .triton folder in the xdg cache directory. That's why I didn't do `export TRITON_HOME="$XDG_CACHE_HOME/triton"` because this would result in `$XDG_CACHE_HOME/triton/.triton`

### DIFF
--- a/programs/triton.json
+++ b/programs/triton.json
@@ -3,7 +3,7 @@
         {
             "path": "$HOME/.triton",
             "movable": true,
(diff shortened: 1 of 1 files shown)


engdyn:
Thinking about it more having `$XDG_CACHE_HOME/triton/.triton` is better because there shouldn't be any hidden folder inside any XDG dirs

b3nj5m1n:
Not ideal but I agree, thank you!

PR #536, opened 2026-01-31
$ELAN_HOME

GrigorenkoPV:
As pointed out in https://github.com/leanprover/elan/issues/75#issuecomment-1627438441


b3nj5m1n:
Thank you!

PR #545, opened 2026-03-20
fix: update python and sqlite history locations

Senderman:
Added description for PYTHON_HISTORY environment variable for user to copy and paste:

```bash
export PYTHON_HISTORY="$XDG_STATE_HOME"/python_history
```

Updated SQLITE_HISTORY file from `"$XDG_CACHE_HOME"/sqlite_history`  to `"$XDG_STATE_HOME"/sqlite_history`, because in all other files like python, postgresql, less etc, history goes to XDG_STATE_HOME


b3nj5m1n:
Thanks!

PR #548, opened 2026-03-31
correct XDG var name in vagrant

haarp:
XDG_DATA_DIR doesn't exist, XDG_DATA_HOME was very ikely intended.


b3nj5m1n:
Thank you!

PR #550, opened 2026-04-09
Update firefox and add SNI and GSConnect

Renarde-dev:
Firefox now support XDG spec but certain programs that use native messaging like GSConnect still create .mozilla.
Also added SNI (https://github.com/alttpo/sni)


b3nj5m1n:
Awesome, thank you!

PR #554, opened 2026-05-04
edit python to fix startup script, backwards-compatibility

fCherkasskiy:
On my systems, the script failed when the python histfile or it's parent directory did not exist on the disk. I used the low-level OS functions to create the directory and the file if needed, rather than relying on readline.write_history_file($PYTHON_HISTORY). Removed walrus operator, as it only works in python>=3.8. Changed PYTHONSTARTUP suggestion to be pythonrc.py to avoid necessity for symlinking. Instead of defining functions, the script starts with a short-circuitable if statement so that in python>=3.13, less overhead. Slightly changed wording/highlighting to improve legibility and typos.


b3nj5m1n:
Thank you!

PR #560, opened 2026-05-27
Added the ability to make an ignore file to avoid unwanted advice

fCherkasskiy:
Added opportunity for the user to create a file in either $XDG_CONFIG_HOME/xdg-ninja/ignore or use an environment variable $XN_IGNOREFILE to list the basenames of files they wish to keep in their $HOME even if a solution exists to move it elsewhere. This is a drop in feature, meaning existing users shouldn't need to change anything if they don't wish to use this feature.

Updated README and MANPAGE

I've been using xdg-ninja for a while, and found I often have files I wish to keep in $HOME such as .zshenv and .xinitrc, and don't like seeing warnings for those. I also wanted to be able to configure this tool without editing the ./programs directory to remove unwanted files. Feel free to let me know if this feature should be disabled by default, but since it doesn't do anything without creating the ignore file, I think it should be enabled by default. Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ Follow the instructions for your package manager to install `xdg-ninja`.
 
 ### Configuration
 
-The configuration is done in the [`./programs/`](./programs/) directory, which should be located in the same working directory as the [`xdg-ninja.sh`](./xdg-ninja.sh) script. This can be overridden with the `XN_PROGRAMS_DIR` environment variable.
+The configuration is done in the [`./programs/`](./programs/) directory, which should be located in the same working directory as the [`xdg-ninja.sh`](./xdg-ninja.sh) script. This can be overridden with the `$XN_PROGRAMS_DIR` environment variable.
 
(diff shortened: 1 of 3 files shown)


lgarron:
Just a drive-by comment: I think `$XDG_NINJA_IGNORE` would be a much better name than `$XN_IGNORE`. It's not much longer, but it's extremely clear what it's about. If I saw that without any further context, I would have no problem figuring out what it's about and what to look up for more info.

fCherkasskiy:
I agree, but I chose `$XN_IGNOREFILE` because of the existing environment variable `$XN_PROGRAMS_DIR`. I wanted to keep the `XN_` prefix to have a similar naming convention, and I don't want to confuse people who are already using `$XN_PROGRAMS_DIR`.
_Side note, the variable is `$XN_IGNOREFILE`, not `$XN_IGNORE`, the commit description had a typo._

fCherkasskiy:
Fixed shellcheck error :)

PR #566, opened 2026-06-05
Add notice about .CFUserTextEncoding on macOS.

scolby33:
This cannot/should not be moved/removed as far as I can tell.

References:
- https://eclecticlight.co/2020/12/10/controlling-processes-and-environments/
- https://developer.apple.com/library/archive/technotes/tn2228/_index.html

### DIFF
--- a/programs/CFUserTextEncoding.json
+++ b/programs/CFUserTextEncoding.json
@@ -4,7 +4,7 @@
         {
             "path": "$HOME/.CFUserTextEncoding",
             "movable": false,
-            "help": "Not applicable. Probably not remvoable.\n\nFrom [Apple](https://developer.apple.com/library/archive/technotes/tn2228/_index.html#//apple_ref/doc/uid/DTS40007991-CH1-SUBSECTION15):\n\n>Core Foundation tries to access the user's home directory to determine their default text encoding (stored in the file ~/.CFUserTextEncoding).\n"
+            "help": "Not applicable. Probably not removable.\n\nFrom [Apple](https://developer.apple.com/library/archive/technotes/tn2228/_index.html#//apple_ref/doc/uid/DTS40007991-CH1-SUBSECTION15):\n\n>Core Foundation tries to access the user's home directory to determine their default text encoding (stored in the file ~/.CFUserTextEncoding).\n"
         }
     ]
 }
--- a/programs/CFUserTextEncoding.json
+++ b/programs/CFUserTextEncoding.json
@@ -1,0 +1,10 @@
+{
+    "name": "macOS Core Foundation user default text encoding",
+    "files": [
+        {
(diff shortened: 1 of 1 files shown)


programs/CFUserTextEncoding.json
```diff
@@ -0,0 +1,10 @@
+{
+    "name": "macOS Core Foundation user default text encoding",
+    "files": [
+        {
+            "path": "$HOME/.CFUserTextEncoding",
+            "movable": false,
+            "help": "Not applicable. Probably not remvoable.\n\nFrom [Apple](https://developer.apple.com/library/archive/technotes/tn2228/_index.html#//apple_ref/doc/uid/DTS40007991-CH1-SUBSECTION15):\n\n>Core Foundation tries to access the user's home directory to determine their default text encoding (stored in the file ~/.CFUserTextEncoding).\n"
```
lgarron:
```suggestion
            "help": "Not applicable. Probably not removable.\n\nFrom [Apple](https://developer.apple.com/library/archive/technotes/tn2228/_index.html#//apple_ref/doc/uid/DTS40007991-CH1-SUBSECTION15):\n\n>Core Foundation tries to access the user's home directory to determine their default text encoding (stored in the file ~/.CFUserTextEncoding).\n"
```
The link says this, though?

> You can prevent this access by setting an environment variable that tells Core Foundation the default text encoding to use.

scolby33: This is true, but I think the warning directly below contradicts this:
> Important: Setting __CF_USER_TEXT_ENCODING may be necessary but not sufficient to ensure that your code runs correctly in this context. Mac OS X's many and various frameworks are littered with references to the user's home directory, and there's no list of what routines are safe to call in this context. Ultimately you will just have to try things and see what works.

It's also something that is checked at early login, and adding keys to the environment there is (I think) not that well supported. (It's not like we can add `export __CF_USER_TEXT_ENCODING` to our bashrc and have it picked up by, for example, Finder.app.)

lgarron: I think it would be really useful to expand the help a bit.

If I read the help from the current PR, I don't think I'd get the right picture without going to the independent link. It certainly isn't unreasonable to leave some details to the link, but I think the help can be a bit more helpful on its own by saying:

> In theory, the original Apple framework using this file also supports using an environment variable instead. However, Apple's own documentation also warns that many other frameworks may only check for the file. Therefore, it is strongly recommended to leave the file in place unless you are willing to risk unexpected text encoding bugs. For more details, see: …

Speaking from my own experience, I am mostly able to replace the file with the env var, *except* that I have to change the UTF-8 encoding mechanism for a few of my scripts.

scolby33: I have updated the help text based on your suggestion :)

